PR #7, opened 2022-02-07
'Solution'

alexandra-halytska:


app/create_file.py
```diff
@@ -1 +1,47 @@
-# write your code here
+import os
+from datetime import datetime
+import argparse
```
MasterpieceElbow, line 3: Don't use `argparse`, use `sys.argv` instead.

MasterpieceElbow (changes requested):

PR #17, opened 2022-06-06
Solution

YuriiKindrat:


app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+def make_directory(path):
+    os.makedirs(path)
```
Flashmobber, line 7: what if path exist?

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+
+def file_creator(name):
+    with open(f"{name}", "a") as f:
+        f.write(datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+        counter = 1
+        while True:
+            content = input("Enter content line: ")
+            if content == "stop":
+                f.write("\n")
```
Flashmobber, line 17: I think it's don't necessary here

Flashmobber (changes requested):
.

app/create_file.py
Flashmobber: @YuriiKindrat , in this line you creates second unnecessary string at the end of file. Ofcourse, it's not critical for this task, but it will be better to remove it.

Flashmobber (changes requested):
read the comments upper, please

Flashmobber (approved):

MasterpieceElbow (approved):

PR #20, opened 2022-06-07
'Solution'

Korenely:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+        path = os.path.join(path, directory)
+    os.makedirs(path)
+    return path
+
+
+def file_generator(file):
+    with open(file, "a") as f:
+        now = datetime.now()
+        f.write(now.strftime("%m-%d-%Y %H:%M:%S") + "\n")
+        i = 1
```
KatiukhaO, line 32: name variable "i" isn`t informative, name this variable relative her responsibility , and code will be more readable

KatiukhaO (changes requested):

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+        while True:
+            content = input("Enter content line: ")
+            if content != "stop":
+                f.write(f"{str(i)} content \n")
+                i += 1
+            else:
+                break
```
MasterpieceElbow, line 39: Bad condition, you can only check for the word `stop` with one `if` statement.

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+def file_generator(file):
+    with open(file, "a") as f:
+        now = datetime.now()
+        f.write(now.strftime("%m-%d-%Y %H:%M:%S") + "\n")
+        i = 1
+        while True:
+            content = input("Enter content line: ")
+            if content != "stop":
+                f.write(f"{str(i)} content \n")
+                i += 1
```
MasterpieceElbow, line 37: Don't name the variable as `i`.

MasterpieceElbow (changes requested):

TPolina (approved):

PR #23, opened 2022-06-07
Solution

Hectorovich:


app/create_file.py
```diff
@@ -1 +1,35 @@
...
+        if not os.path.exists(path):
+            os.makedirs(path)
+    elif "-f" in command and "-d" not in command:
+        create_file(path, command[-1])
+
+
+def create_file(path, name):
+    with open(path + "\\" + name, "a+") as file:
+        file.write(datetime.datetime.now().strftime("%m-%d-%Y, %H:%M:%S\n"))
+        while True:
```
KatiukhaO, line 27: in example for this task, shown, that, every line will mast have number, but , as, I see, that, in your code isn`t this

KatiukhaO (changes requested):

KatiukhaO (approved):

PR #24, opened 2022-06-07
Solution

darten360:


rotsen18 (changes requested):
Main problem is that your app woun't work from terminal. There isn't func call inside.

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+                    break
+                f.write(f"{count} {input_string}\n")
+                count += 1
+
+    if "-d" in sys.argv and "-f" in sys.argv:
+        create_full_path(os.path.sep.join(
+            sys.argv[sys.argv.index("-d") + 1:sys.argv.index("-f")]),
+            sys.argv[-1])
+    elif "-d" in sys.argv:
+        os.makedirs(
```
rotsen18, line 28: Here you do not check if such a path already exists, although in lines 9-10 there is such a check. If you add the same check here, you will not follow the DRY principle. It may be best to divide the functions more carefully into folder and file creation, but that's up to you. And try to not use f,i,k,v for variables name.

rotsen18 (approved):

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+
+def create_file():
+    def create_full_path(path, name):
+        print(path)
+        if not os.path.exists(path):
+            os.makedirs(path)
+
+        with open(path + os.path.sep + name, "a") as f:
+            f.write(datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+            count = 1
```
TPolina, line 14: where do you need this variable?

app/create_file.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import datetime
+import os
+import sys
+
+
+def create_file():
+    def create_full_path(path, name):
```
TPolina, line 7: take `create_full_path` function outside `create_file` function.

app/create_file.py
TPolina: I agree, better write separate function `create_dirs` and use it

PR #25, opened 2022-06-07
solution

mens0:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+        file.write(f"{now}\n")
+        while True:
+            message = input("Enter content line: ")
+            if message == "stop":
+                break
+            file.writelines(f"{counter} {message}\n")
+            counter += 1
+
+
+input_data = sys.argv
```
KatiukhaO, line 30: i think, from $30 - 44 lines code, will be better to separate in separate function

KatiukhaO (changes requested):

MasterpieceElbow (approved):

PR #28, opened 2022-06-07
Solution

RomaTsybulia:


app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
```
Serhii-Leonenko, line 4: I don't exactly know can program be created without fuctions, but I think it's not good practice, and it is realy unreadability.

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+                break
+
+            dir_path = os.path.join(dir_path, arguments[j])
+            os.mkdir(dir_path)
+
+    elif arguments[i - 1] == "-f":
+        line_number = 1
+
+        if os.path.exists(arguments[i]):
+            with open(arguments[i], "a") as file:
```
Serhii-Leonenko, line 28: You should use descriptive names for variables. Like example ,``with open(file_name, "a") as file` `  is better `with open(arguments[i], "a") as file`

Serhii-Leonenko, line 28: You should use descriptive names for variables. Like example ,with open(file_name, "a") as file is better than with open(arguments[i], "a") as file.

So, i recommend you to devide your file on fuctions, and use more descriptive names

Serhii-Leonenko (changes requested):

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+                os.mkdir(dir_path)
+            break
+
+        elif arguments[i - 1] == "-f":
+            file_name = os.path.join(dir_path, arguments[i])
+            create_file(file_name)
+            break
+
+
+create_file_and_directories()
```
Serhii-Leonenko, line 46: you can add 
if __name__ == "__main__":

Serhii-Leonenko (approved):

TPolina (approved):

PR #29, opened 2022-06-07
Solution

BohdanKryven:


Chardona (changes requested):
What if you have -d and then -f? You haven't handled that case.

PR #36, opened 2022-06-07
'Solution'

SergeyDev1996:


app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+from datetime import datetime
+import os
+import sys
+
+parent_dir = r"C:\PytyhonLearn\py-create-file-from-terminal\app"
```
Serhii-Leonenko, line 5: if you run it  on another computer, will it work?
use `os.getcwd()`

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+time = datetime.now()
+timee = time.strftime("%Y/%m/%d %H:%M:%S")
+
+if "-f" in sys.argv and "-d" in sys.argv:
+    copy = sys.argv
+    d_index = copy.index("-d")
+    f_index = copy.index("-f")
+    dir_ = copy[d_index + 1:f_index]
+    path = '/'.join(dir_)
```
Serhii-Leonenko, line 16: think about that to `use os.path.join()`

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+elif "-f" in sys.argv:
+    name = sys.argv[2]
+    current_name = name
+    with open(current_name, 'a') as f:
+        f.write(str(timee) + "\n")
+        count = 1
+        while True:
+            content = input('Enter the line content:')
+            if content == 'stop':
+                break
```
Serhii-Leonenko, line 46: Does your fuction create empty string like this?
image

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+
+time_now = datetime.now()
+time_formated = time_now.strftime("%Y/%m/%d %H:%M:%S")
+
+if "-f" in sys.argv and "-d" in sys.argv:
+    copy = sys.argv
+    d_index = copy.index("-d")
+    f_index = copy.index("-f")
+    directory = copy[d_index + 1:f_index]
+    directoryy = str(directory)
```
TPolina, line 16: where do you need this variable?

TPolina (changes requested):
you have duplicated blocks of code. Write separate function for creating directories and function for creating files. Then use them in main function with the main logic

PR #38, opened 2022-06-08
Solution

Evgeniy994:


app/create_file.py
```diff
@@ -1 +1,39 @@
...
+def check_directory(directory):
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+    return directory
+
+
+def create_input(directory, file_name):
+    with open(directory + file_name, "a") as f:
+        date_now = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        f.write(date_now + "\n")
```
rotsen18, line 16: You will have newline after date string but in task newline is before.

rotsen18 (changes requested):
Your app won't work from terminal. You should to call main func isnside.

rotsen18 (approved):

TPolina (approved):

PR #41, opened 2022-06-08
Solution

vasylhnatiuk:


rotsen18 (changes requested):
You didn't call function in module so your code won't work from terminal and try to use more descriptive names for variables (f -> file for example)

rotsen18 (approved):

TPolina (approved):

PR #62, opened 2022-07-31
Solution

andriy-demeshko:


IvanRamyk (approved):
Good job! 
But it'd be better to create a function that does all this logic and then execute it. It's a better solution because in that case, it's easy to reuse the code.

PR #68, opened 2022-08-17
Solutions

InnaKushnir:


DmytroSvirsa (changes requested):
make a function, please

app/create_file.py
```diff
@@ -2,34 +2,38 @@
...
+    with open(name_file, "a") as f:
+        f.write(str(datetime.now())[:-7] + "\n")
+        print(str(datetime.now())[:-7] + "\n")
+        i = 0
+        print("Enter content line: ")
+        for line in sys.stdin:
+            i += 1
+            if "stop" == line.rstrip():
+                break
+            else:
```
DmytroSvirsa, line 36: You don’t need `else` - you have a `break` after condition. Rewrite without else

DmytroSvirsa (changes requested):

app/create_file.py
```diff
@@ -33,7 +33,6 @@ def create_file():
             i += 1
             if "stop" == line.rstrip():
                 break
-            else:
-                f.write(f"{i} Line{i} {line}")
-            print(f"Enter content line:"
-                  f" Line{i} {line}")
+
+    f.write(f"{i} Line{i} {line}")
+    print(f"Enter content line: Line{i} {line}")
```
DmytroSvirsa, line 38: but it still must be inside your `for` loop correct indentation

DmytroSvirsa (changes requested):

app/create_file.py
InnaKushnir: My bad

DmytroSvirsa (approved):

PR #77, opened 2022-09-01
Solution

anastasia-martyniuk:


app/create_file.py
```diff
@@ -1 +1,75 @@
...
+def make_directory_and_file():
+    parent_dir = os.getcwd()
+    for i in text_from_terminal[d_index + 1:f_index]:
+        directory = str(i)
+        path = os.path.join(parent_dir, directory)
+        parent_dir = path
+    os.makedirs(parent_dir)
+
+    new_path = os.path.join(parent_dir, text_from_terminal[-1])
+    with open(new_path, "a") as f:
```
IvanRamyk, line 52: Looks like a duplication
Avoid such situations

IvanRamyk (changes requested):

PR #85, opened 2022-09-16
Solution

VladyslavKlevanskyi:


app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        path_file_name = file_name
+    else:
+        path_file_name = f"{path}\\{file_name}"
+
+    try:
+        os.makedirs(path)
+    except FileExistsError:
+        pass
+    except FileNotFoundError:
+        pass
```
Abnormaltype, line 32: nad practice, don't use pass in the exception

Abnormaltype (changes requested):

app/create_file.py
VladyslavKlevanskyi: What should I use in exeption, if don't want that my function do anything in exception?

Abnormaltype: create a function or functions for this action

Abnormaltype (approved):
Well done!

PR #86, opened 2022-09-18
Solution for creat file

Foggy-Cat:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+from sys import argv
+from os import makedirs
+from datetime import datetime
+
+
+def make_dir(paths: list):
+    try:
+        d_index = paths.index("-d")
+    except ValueError:
+        return
```
IvanRamyk, line 10: maybe `print` message to display the error type on screen

IvanRamyk (changes requested):

app/create_file.py
Foggy-Cat: but we can just create file, without create dir. it wasn't error, return for stoped function

IvanRamyk: agree

IvanRamyk (approved):
Good job!

PR #87, opened 2022-09-20
Solution

KatyaVasylieva:


IvanRamyk (changes requested):
it's better to create a main function that performs all actions and than call it from `if __name__ == "__main__" block`

IvanRamyk (approved):
Good job!

PR #100, opened 2022-09-29
Solution

AVyha:


app/create_file.py
```diff
@@ -1 +1,25 @@
...
+
+
+directory = ""
+command_list = sys.argv
+if command_list[1] == "-d":
+    try:
+        f_command = command_list.index("-f")
+    except ValueError:
+        f_command = None
+    os.makedirs("\\".join(command_list[2:f_command]))
```
mltdll, line 13: To make your solution cross-platform, you can use `os.path.join()` function. However, it takes names as arguments, not an iterable, so you'll have to unpack them.
Also, make use of `exists_ok=True` flag for `os.makedirs`. That way it won't throw an error when directory already exists.

app/create_file.py
```diff
@@ -1 +1,25 @@
...
+    directory = "\\".join(command_list[2:f_command]) + "\\"
```
mltdll, line 14: Use `os.path.join()` here as well, and add `os.path.sep` instead of hardcoded `\`.

mltdll (changes requested):

mltdll (approved):
Looks okay.

app/create_file.py
```diff
@@ -1 +1,29 @@
...
+    except ValueError:
+        f_command = None
+    directory = os.path.join(*command_list[2:f_command])
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+
+if "-f" in command_list:
+    f_index = command_list.index('-f')
+    with open(f"{os.path.join(directory, command_list[f_index + 1])}",
+              "a") as f:
```
Nattalli, line 20: as `f` is a bad name

app/create_file.py
```diff
@@ -1 +1,29 @@
...
+        f_command = command_list.index("-f")
+    except ValueError:
+        f_command = None
+    directory = os.path.join(*command_list[2:f_command])
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+
+if "-f" in command_list:
+    f_index = command_list.index('-f')
+    with open(f"{os.path.join(directory, command_list[f_index + 1])}",
```
Nattalli, line 19: maybe you can just call `with open(str(...))`?

Nattalli (changes requested):

app/create_file.py
mltdll: But isn't `os.path.join()`'s output already a string? Why convert it again?

Nattalli: that's a good question. But I thought that if @AVyha use `f-string` that means that it doesn't work correctly without it

Nattalli (approved):
Good job! 😸

PR #101, opened 2022-09-29
'Solution'

davidkrivko:


DmytroSvirsa (changes requested):
let's improve your code style. let's protect it from running everything from another module. Put everything into a function, and call it only when this module run

DmytroSvirsa (approved):

PR #102, opened 2022-10-02
Solution

luxuriant777:


app/create_file.py
```diff
@@ -1 +1,37 @@
...
+        line_number = 1
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        while True:
+            line = input("Enter content line: ")
+            if line.lower() == "stop":
+                break
+            file.write(f"{line_number} {line}\n")
+            line_number += 1
+
+    def create_file_from_terminal() -> None:
```
TPolina, line 22: why did you write one function inside another?

TPolina (changes requested):

app/create_file.py
luxuriant777: refactored

TPolina (approved):
Good job!

PR #106, opened 2022-10-04
Solution

taras-andruschenko:


kateryna3k (approved):
Looks nice!

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+    commands = {
+        "-d": None,
+        "-f": None,
+    }
+    for i, char in enumerate(sys.argv[1:]):
+        if char in commands:
+            while i + 1 >= len(sys.argv) or sys.argv[i + 1] in commands:
+                commands[char] = sys.argv[i + 2]
+                i += 1
+    return commands.get(key)
```
TPolina, line 17: I think this function is a bit complicated and you call it too many times in your code. You can just search for '-d' and '-f' in command and if they are there use indexes to get file and directories. For example, `file_name = command[-1]`. Than this function can return file_name and directories so you don't need to call it many times.

TPolina (changes requested):

app/create_file.py
taras-andruschenko: if you have more than two flags.. much more, what will you do in this case?)

TPolina (approved):
Good job!

PR #107, opened 2022-10-04
Solution

pazyn404:


app/create_file.py
```diff
@@ -1 +1,34 @@
...
+
+filename = ""
+directories = []
+
+if "-f" in argv:
+    filename = argv[argv.index("-f") + 1]
+    if "-d" in argv:
+        directories = argv[argv.index("-d") + 1:argv.index("-f")]
+else:
+    filename = "file.txt"
```
mltdll, line 14: I don't think it follows from the task instructions that you should implicitly write to `file.txt` if the filename is absent. My understanding is that you shouldn't create file at all, only directories.

mltdll (changes requested):

mltdll (approved):

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+    except FileExistsError:
+        pass
+    chdir(directory)
+
+if filename:
+    file_data = [datetime.today().strftime("%y-%m-%d %H:%M:%S"), "\n"]
+
+    while True:
+        line = input("Enter content line: ")
+        if line == "exit":
```
Y-Havryliv, line 29: it should be `stop` not `exit`

Y-Havryliv (changes requested):

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+            break
+        file_data.extend([line, "\n"])
```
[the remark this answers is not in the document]
pazyn404: Yes, I test it. So I fixed "exit" and add line number to output. I can't do changes on this branch so I create new one.

PR #111, opened 2022-10-05
Solution

shurushku:


mltdll (approved):
Looks good to me.

app/create_file.py
```diff
@@ -1 +1,35 @@
...
+            i += 1
+
+
+if __name__ == '__main__':
+    if len(argv) > 2:
+        param = get_commands([arg.strip() for arg in argv])
+        if param:
+            if param["path"] and not path.isdir(param["path"]):
+                makedirs(param["path"])
+            write_to_file(path.join(param["path"], param["file_name"]))
```
TPolina, line 35: it's better to create one more function ('main' or 'create_file' for example) and put this code there. And in `if __name__ == '__main__':` only run this function.

app/create_file.py
```diff
@@ -1 +1,35 @@
-# write your code here
+from sys import argv
+from os import path, makedirs
+from datetime import datetime
+
+
+def get_commands(commands: list) -> dict[str] | None:
+    if "-f" not in commands:
+        return
```
TPolina, line 8: What if `-f` is not in commands but `-d` in commands? You need to create directories in this case as well

TPolina (changes requested):

TPolina (approved):
Good job!

PR #112, opened 2022-10-05
Solution

Andriy-Makarenko:


app/create_file.py
```diff
@@ -1 +1,49 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def execute() -> None:
```
mltdll, line 6: I can see your idea here, but is this function necessary when all it does is just one function call? And isn't there a convention to call a function like this `main`?
This is a question where I'd rather get an answer and not just change in code. :)

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+
+
+def new_directory(path: str) -> None:
+    os.makedirs(path, exist_ok=True) if path else None
+
+
+def create_a_file(file_name: str) -> None:
+    with open(file_name, "a") as file_out:
+        line_number = 1
+        time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
mltdll, line 17: Time shadows a built-in module name. I'd suggest to pick another name.

mltdll (changes requested):

app/create_file.py
Andriy-Makarenko: Done

mltdll (approved):

app/create_file.py
```diff
@@ -1 +1,49 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def main() -> None:
+    get_path_from_terminal()
```
TPolina, line 7: why do you need this function? You can run `get_path_from_terminal` in 
`if __name__ == "__main__":` as well

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+
+        while user_input.lower() != "stop":
+            file_out.write(f"{line_number} {user_input}\n")
+            line_number += 1
+            user_input = input("Enter content line: ")
+
+        file_out.write("\n")
+
+
+def get_path_from_terminal() -> None:
```
TPolina, line 29: it's better to rename this function because it actually not only gets path from terminal but also creates directories and file

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        new_directory(path)
+        file_name = "/".join([path, file_name])
+        create_a_file(file_name)
+    elif "-d" in command:
+        index_d = command.index("-d")
+        path = "/".join(command[(index_d + 1):])
+        new_directory(path)
+    elif "-f" in command:
+        file_name = command[-1]
+        create_a_file(file_name)
```
TPolina, line 45: it's ok, but you have code duplication here. You can try to rewrite this function. You can do only two checks: `if -f in command` and `if -d in command` and use the fact that `-f` flag and file_name took 2 last places in a list of arguments. So one `if` block will be an addition to another. I can give you an example if you will try and will need help

TPolina (changes requested):

app/create_file.py
Andriy-Makarenko: You told me on Q&A it's a good practise to have a separate function to call other functions

TPolina: I don't remember exactly what I said, but it seems to me that I said that's a good practice to have `if __name__ == "__main__":` and call one function inside it which will call other functions and combine logic from them

TPolina (approved):
It's not actually what I meant when proposed to try to simplify getting file_name and directories, but ok. You could do sth like this:
```
if "-f" in command:
    file_name = command[-1]
    command = command[:-2]
if "-d" in command:
    directories = command[1:]
```

PR #114, opened 2022-10-05
Solution

Aleksandro777:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+        create_file(file_name)
+    if "-d" in command:
+        index_d = command.index("-d")
+        file_path = command[index_d + 1:]
+        create_directory(file_path)
+
+
+def create_directory(path_list: list) -> None:
+    path = "/".join(path_list)
+    os.makedirs(path)
```
mltdll, line 28: Does it work if you try to create a file in an existing directory?
I'd suggest to either check for directory existence or using `exists_ok` argument of `makedirs()`.

mltdll (changes requested):

TPolina (approved):

PR #116, opened 2022-10-05
Solution)

Gr1zzZzzly:


app/create_file.py
```diff
@@ -1 +1,28 @@
...
+import os
+import datetime
+
+
+def create_file() -> None:
+    enter = sys.argv
+    path = ""
+
+    if "-d" in enter:
+        path = os.path.join(*enter[(enter.index("-d") + 1):enter.index("-f")])
```
TPolina, line 11: a bit complicated. `-f` is always in the -2 place and so on

app/create_file.py
```diff
@@ -1 +1,28 @@
...
+        if os.path.exists(rf"{path}\{enter[-1]}") is False:
```
TPolina, line 12: `is False` is unnecessary

app/create_file.py
```diff
@@ -1 +1,28 @@
...
+            os.makedirs(path)
+
+    with open(rf"{path}\{enter[-1]}", "a") as f:
```
TPolina, line 15: what if `-f` is not in the command

TPolina (changes requested):

app/create_file.py
```diff
@@ -1 +1,32 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+def create_file() -> None:
+    enter = sys.argv
+    print(enter)
```
TPolina, line 8: unnecessary print

TPolina (approved):
Good job!

app/create_file.py
Gr1zzZzzly: oh sry

PR #117, opened 2022-10-06
Solution

linaSam:


hasslieb3 (approved):

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+
+        count = 0
+
+        while True:
+            text_input = input("Enter content line: ")
+            count += 1
+            if text_input == "stop":
+                break
+            else:
+                result.write(f"{count} {text_input}\n")
```
Y-Havryliv, line 31: did you test your code?

Y-Havryliv (changes requested):

app/create_file.py
linaSam: not really - didn't understand how. already rewrote code and tested it properly

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+    if path:
+        os.makedirs("/".join(path))
+        file_name = f"{'/'.join(path)}/{file_name}"
+    add_information_to_file(file_name)
+
+
+def directory_file_creation() -> None:
+    entered_info = sys.argv
+
+    if "-d" in entered_info and "-f" not in entered_info:
```
Nattalli, line 32: trying call `python create_file.py -d dir1 dir2` an error occurs:

os.makedirs("/".join(*entered_info[(entered_info.index("-d") + 1):]))
TypeError: str.join() takes exactly one argument (2 given)

Nattalli (changes requested):

app/create_file.py
Nattalli: now it creates a such directory 
Снимок экрана 2022-10-11 в 11 54 38

app/create_file.py
```diff
@@ -1 +1,47 @@
-# write your code here
+import datetime
+import os
+import sys
+
+
+def add_information_to_file(file_name: str) -> None:
+    with open(file_name, "a") as document:
```
Nattalli, line 7: In the examples of the task when inserting new text, first comes the empty line, and then the new values

Nattalli (changes requested):

Nattalli (changes requested):

Nattalli (approved):
You got it! Well done! 👍🏻

PR #119, opened 2022-10-06
Solution

rostyak:


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+import os
+from datetime import datetime
+
+
+def create_file() -> None:
+    os.chdir("app")
+    if "-d" in sys.argv:
+        new_path = []
+        for command in sys.argv[sys.argv.index("-d") + 1:]:
+            print(command)
```
VadymShkarbul, line 11: Maybe this print is no longer needed ;)

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+            if command == "-f":
+                break
+            else:
+                new_path.append(command)
+        os.makedirs(os.path.join(*new_path))
```
VadymShkarbul, line 16: exist_ok=True to specify an existing directory without error

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+        with open(sys.argv[sys.argv.index("-f") + 1], "a") as new_file:
+            current_datetime = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            new_file.write(current_datetime + "\n")
+            line = 1
+            while True:
+                string = input("Enter content line: ")
+                if string == "stop":
+                    new_file.write("\n")
+                    break
+                new_file.write(str(line) + " " + string + "\n")
```
VadymShkarbul, line 29: as for me f string looks better, but up to you

app/create_file.py
```diff
@@ -1 +1,33 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_file() -> None:
+    os.chdir("app")
```
VadymShkarbul, line 7: It's good idea to use chdir in the end of this function, but as for me, it's not nessesary at the beginning, because main.py is already in the app. Maybe I didn't understand something. Please tell me about it.

VadymShkarbul (changes requested):

app/create_file.py
rostyak: It doesn't do anything. I forgot to remove it, my bad.

VadymShkarbul (approved):
GJ!

Nattalli (approved):
Good job!

PR #121, opened 2022-10-06
Solution

HStakhiv:


Yaroslav-Stetsko (approved):
Great gob!

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+            if user_input == "stop":
+                file.write("\n")
+                break
+
+            line_counter += 1
+            file.write(f"{line_counter} {user_input}\n")
+
+
+if __name__ == "__main__":
+    get_from_terminal()
```
Y-Havryliv, line 48: try to test youd code when creating multiple directories with `-d` flag

Y-Havryliv (changes requested):

app/create_file.py
HStakhiv: Like this `python app/create_file.py -d dir1 dir2 dir3 d4 d5 d6 `?

HStakhiv: That's right, I lost one folder ((

Y-Havryliv: all good, nice job

HStakhiv: Thanks

Y-Havryliv (approved):
GJ!

PR #122, opened 2022-10-07
Solution

pazyn404:


app/create_file.py
```diff
@@ -1 +1,35 @@
...
+    file_data = [f"{datetime.today().strftime('%y-%m-%d %H:%M:%S')}\n"]
+    line_num = 0
+    while True:
+        line_num += 1
+        line = input("Enter content line: ")
+        if line == "stop":
+            break
+        file_data.extend(f"{line_num} {line}\n")
+
+    with open(filename, "w") as file:
```
Nattalli, line 34: From the task:
`If file.txt already exists it should add content below:`
```python create_file.py -f file.txt```

Nattalli (changes requested):

PR #123, opened 2022-10-07
Solution

skrypka-dmytro:


app/create_file.py
```diff
@@ -1 +1,30 @@
...
+
+    if "-d" in info:
+        path += os.path.join(
+            *info[(info.index("-d") + 1):info.index("-f")]
+        )
+        if os.path.exists(rf"{path}\{info[-1]}") is False:
+            os.makedirs(path)
+
+    with open(rf"{path}\{info[-1]}", "a") as file:
+        data_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
mltdll, line 18: It should be `date_time`, or `current_time`, or something like this. "data" doesn't really fit in.

app/create_file.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def file_make() -> None:
```
mltdll, line 6: Why not `create_file`? Functions are usually called this way.

app/create_file.py
```diff
@@ -1 +1,30 @@
...
+    info = sys.argv
+    path = ""
+
+    if "-d" in info:
+        path += os.path.join(
+            *info[(info.index("-d") + 1):info.index("-f")]
+        )
+        if os.path.exists(rf"{path}\{info[-1]}") is False:
```
mltdll, line 14: Since we can be sure that `os.path.exists()` always returns a bool value, you can rewrite it as `if not ...` instead of `if ... is False`.

mltdll (changes requested):

mltdll (approved):

app/create_file.py
```diff
@@ -1 +1,30 @@
...
+from datetime import datetime
+
+
+def create_file() -> None:
+    info = sys.argv
+    path = ""
+
+    if "-d" in info:
+        path += os.path.join(
+            *info[(info.index("-d") + 1):info.index("-f")]
```
Nattalli, line 12: did you test your code?
calling `python create_file.py -d dir1 dir2` the program returns:
 ```python
*info[(info.index("-d") + 1):info.index("-f")]
ValueError: '-f' is not in list
```

**Review by Nattalli — CHANGES_REQUESTED** [review]

**Review by Nattalli — CHANGES_REQUESTED** [review]
Please, check your code again. Now it does not execute any commands correctly

**Review by Nattalli — CHANGES_REQUESTED** [review]
Well, 2 more things:
1) `python create_file.py -f file.txt` doesn't create new file
2) `python create_file.py -d dir1 dir2 -f file.txt` return an error if dir1 dir2 already exist:
```python
FileExistsError: [Errno 17] File exists: 'dir1/dir2'
```

Nattalli (approved):
👍🏻

PR #129, opened 2022-10-07
Solution

graandda:


HStakhiv (changes requested):
The program does not work if the folders already exist. I can't add a file to an existing folder

HStakhiv (approved):
Now everything works. Good

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+def create_path(path: str) -> None:
+    try:
+        os.makedirs(path)
+
+    except Exception:
+        pass
+
+
+def create_file(name: str, path: str = "") -> None:
+    with open(path + name, "w") as file:
```
Nattalli, line 33: it is a wrong mode:
```
The app should add the current timestamp at the top and number lines. If file.txt already exists it should add content below
```

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+        time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{time}\n")
+        line_number = 1
+        content_line = input("Enter content line: ")
+        while content_line != "stop":
+            file.write(f"{line_number} Line{line_number} {content_line}\n")
```
Nattalli, line 39: it is incorrect text, should be something like this:
```python
file.write(f"{line_number} {content_line}\n")
```

Nattalli: without `Line` word

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+        pass
+
+
+def create_file(name: str, path: str = "") -> None:
+    with open(path + name, "w") as file:
+        time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{time}\n")
+        line_number = 1
+        content_line = input("Enter content line: ")
+        while content_line != "stop":
```
Nattalli, line 38: also, don't forget add blank line if comand == "stop":
```
In the examples of the task when inserting new text, first comes the empty line, and then the new values
```

Nattalli (changes requested):

app/create_file.py
graandda: tnx)

Nattalli (approved):
Awesome!

PR #130, opened 2022-10-07
Solved the task

Cerne13:


app/create_file.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_file() -> None:
+    if "-d" in sys.argv:
+        i = 2
+        while i < len(sys.argv) and sys.argv[i] != "-f":
```
mltdll, line 9: You can use `os.makedirs()` to simplify this function a bit. This function also has an optional parameter that will be useful here.

mltdll (changes requested):
This solution can look cleaner if you use `argparse` module.

By the way, I guess, all the changes I suggested are optional. Rejecting just to make sure you'll see.

mltdll (approved):

MasterpieceElbow (approved):
GJ!

PR #131, opened 2022-10-08
'Solution'

Menselface:


app/create_file.py
```diff
@@ -1 +1,37 @@
...
+)
+
+args = parser.parse_args()
+
+create_file = args.f
+create_folders = args.d
+timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+if create_folders:
+    os.makedirs("/".join(args.d))
```
mltdll, line 25: It's better to use `os.path.join()` here.

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+
+if create_file:
+    count_lines = 1
+    with open(os.path.join("/".join(create_folders), create_file), "a") as f:
```
mltdll, line 29: Using `join()` is completely unnecessary here. Just unpack the variable:
```python
os.path.join(*create_folders, create_file), "a")

File: app/create_file.py [hunk 3]
```diff
@@ -1 +1,37 @@
...
+        f.write(f"{timestamp}\n")
```
mltdll, line 30: I'd add `"\n"` to the timestamp directly in `strftime` method. This way you wouldn't have to use an f-string here.
By the way, why do you even create timestamp so early? Wouldn't it be better to create it directly above this line?

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+)
+parser.add_argument(
+    "-f",
+    type=str,
+    help="Create file",
+)
+
+args = parser.parse_args()
+
+create_file = args.f
```
mltdll, line 20: Variable names are usually nouns, not verbs. It would be better to call these two something like "file/folders_to_create".

mltdll (changes requested):

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+
+if create_folders:
+    os.makedirs("/".join(args.d))
+
+if create_file:
+    count_lines = 1
+    with open(os.path.join("/".join(create_folders), create_file), "a") as f:
+        f.write(f"{timestamp}\n")
+
+        while True:
```
mltdll, line 32: You can also rewrite this loop as follows:
```python
while (line := input("Enter content line:")) != "stop":
```
And get rid of the variable assignment and the check inside.

mltdll (approved):
Awesome!

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+    os.makedirs(os.path.join(*folders))
+
+if file_name:
+    count_lines = 1
+    with open(os.path.join(*folders, file_name), "a") as f:
+        f.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+
+        while (line := input("Enter content line:")) != "stop":
+            f.write(f"{count_lines} {line}\n")
+            count_lines += 1
```
Nattalli, line 33: add blank line at the end of file

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+
+file_name = args.f
+folders = args.d
+
+if folders:
+    os.makedirs(os.path.join(*folders))
+
+if file_name:
+    count_lines = 1
+    with open(os.path.join(*folders, file_name), "a") as f:
```
Nattalli, line 28: In the examples of the task when inserting new text, first comes the empty line, and then the new values

Nattalli (approved):

PR #132, opened 2022-10-08
App for creating directories and file with appending from terminal

alexzubkoff:


app/create_file.py
```diff
@@ -1 +1,57 @@
...
+
+def create_file_from_terminal() -> None:
+    commands_dict = {}
+    if "-d" in sys.argv[1:] and "-f" not in sys.argv[1:]:
+        if "-" not in sys.argv[1:]:
+            commands_dict["-d"] = sys.argv[1:][sys.argv[1:].index("-d") + 1:]
+        else:
+            commands_dict["-d"] = sys.argv[1:][sys.argv[1:].index("-d") + 1:
+                                               sys.argv[1:].index("-")]
+    if "-d" in sys.argv[1:] and "-f" in sys.argv[1:]:
```
serviza, line 28: I think this can be simplified a little. 
If you check condition with '-d' and '-f' first
You can use 'elif' and don't write 'not in sys.argv[1:]' again and again.

serviza (changes requested):

app/create_file.py
alexzubkoff: Agree! But if i will remove that:
        if "-" not in sys.argv[1:]:
            commands_dict["-d"] = sys.argv[1:][sys.argv[1:].index("-d") + 1:]
        else:
            commands_dict["-d"] = sys.argv[1:][sys.argv[1:].index("-d") + 1:
                                               sys.argv[1:].index("-")]
 When user enter from terminal such as:  
 python  create_file.py  -d   dir1 dir2  -  file.txt  
 i will create directory such as:  dir1/dir2/-/file.txt    
I think it will not be quite right, although the systems such Windows and Linux  will allow it...

serviza (approved):

app/create_file.py
serviza: I meant below code after line #28

alexzubkoff: If i will do with "else" ,  all cases  such "python create_file.py -  dir1 dir 2 - file.txt" will fall there. But I won't process them at all. Because they don't have full-fledged flags such -f and -d.

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+        else:
+            create_file_or_append_to_existing(f"{dir_path}/{file_name}", "a")
+    if file_name and not dir_path:
+        if not os.path.isfile(file_name):
+            create_file_or_append_to_existing(file_name, "w")
+        else:
+            create_file_or_append_to_existing(file_name, "a")
+
+
+create_file_from_terminal()
```
Nattalli, line 57: add blank line at the end of file

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+            os.makedirs(dir_path)
+        if not os.path.isfile(f"{dir_path}/{file_name}"):
+            create_file_or_append_to_existing(f"{dir_path}/{file_name}", "w")
+        else:
+            create_file_or_append_to_existing(f"{dir_path}/{file_name}", "a")
+    if file_name and not dir_path:
+        if not os.path.isfile(file_name):
+            create_file_or_append_to_existing(file_name, "w")
+        else:
+            create_file_or_append_to_existing(file_name, "a")
```
Nattalli, line 54: In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+    count = 1
+    my_file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+    while True:
+        input_str = str(input("Enter content line: "))
+        if input_str != "stop":
+            my_file.write(f"{count} {input_str}\n")
+        else:
+            break
+        count += 1
+    my_file.close()
```
Nattalli, line 17: doesn't it more simple to use `with open() as file`...?

app/create_file.py
Nattalli, line 54: can't a logic with `w` and `a` modes be realized just with `a` mode?

Nattalli (changes requested):

app/create_file.py
```diff
@@ -1 +1,51 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_file_or_append_to_existing(file_name: str, mode: str) -> None:
+    with open(file_name, mode) as my_file:
+        count = 1
+        my_file.write("\n")
```
Nattalli, line 9: notice: now it creates a blank line at the start of the file:
Снимок экрана 2022-10-11 в 13 37 16

Better add it to the 16 rows. When the command is equal to `stop`, you will add a blank line.

Nattalli (changes requested):
Left one comment, everything else works fine!

app/create_file.py
alexzubkoff: Thanks! Sure that is better.

PR #133, opened 2022-10-08
solution

serviza:


alexzubkoff (approved):
Good job! You have  only 33 rows of code! It is a cool!
But how it works, let mentors check. 
But i think if you create one function you can create another for example "create_file_from_terminal" for pack another code for better reading.

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+
+if ("-d" and "-f") in sys.argv:
+    arg_d = sys.argv.index("-d")
+    arg_f = sys.argv.index("-f")
+    file_name = sys.argv[arg_f + 1]
+    directory = sys.argv[arg_d + 1: arg_f]
+    os.makedirs("/".join(directory))
+    file_name_with_dir = "/".join(directory) + "/" + file_name
+    create_file(file_name_with_dir)
+elif "-f" in sys.argv:
```
Nattalli, line 30: did you test your code? Trying call `python create_file.py -f file.txt` will occur the next:
```
 arg_d = sys.argv.index("-d")
ValueError: '-d' is not in list
```

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+        f.write(f"{time}\n")
+        line_num = 1
+        line = input("Enter content line: ")
+        while line != "stop":
+            f.write(f"{line_num} Line{line_num} {line}\n")
+            line = input("Enter content line: ")
+            line_num += 1
+
+
+if ("-d" and "-f") in sys.argv:
```
Nattalli, line 22: also, trying add a file with existed way will occur:
```
FileExistsError: [Errno 17] File exists: 'dir1/dir2'
```
But it should not work in this way

Nattalli (changes requested):

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+    file_name_with_dir = "/".join(directory) + "/" + file_name
+    create_file(file_name_with_dir)
+elif "-f" in sys.argv:
+    arg_f = sys.argv.index("-f")
+    file_name = sys.argv[arg_f + 1]
+    create_file(file_name)
+elif "-d" in sys.argv:
+    arg_d = sys.argv.index("-d")
+    directory = sys.argv[arg_d + 1:]
+    os.makedirs("/".join(directory))
```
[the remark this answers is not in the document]
serviza: But in pycharm there is a line

app/create_file.py
serviza: Yes, I found the mistake)

Nattalli: did you fix it? I still have the same error 😢

Nattalli (changes requested):
Fix next moments:
1) calling `python create_file.py -d dir1 dir2 -f file.txt` if dir1 dir2 already exist
2) add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
serviza: I didn't understand an isuue the first time. Sorry)

Nattalli (approved):
Good job!

PR #136, opened 2022-10-09
'Solution'

OvDima1:


app/create_file.py
```diff
@@ -1 +1,29 @@
...
+import os
+from datetime import datetime
+
+
+def create_file() -> None:
+    if "-d" in sys.argv:
+
+        index = 2
+
+        while len(sys.argv) != index and sys.argv[index] != "-f":
```
mltdll, line 11: You can simplify this functionality if you use `os.makedirs()`.
Hint: if you do, please make sure to check what optional parameters this function has. One of them will be useful.

mltdll (commented):
If you want, you can make this function look cleaner by using `argparse` module.

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+        for commands in command[command.index("-d") + 1:]:
+            if commands == "-f":
+                break
+            else:
+                path_new.append(commands)
+        os.makedirs(os.path.join(*path_new), exist_ok=True)
+        os.chdir(os.path.join(*path_new))
+
+    if "-f" in command:
+        with open(command[-1], "w") as file:
```
Nattalli, line 20: should be "a" mode

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+            date = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            file.write(f"{date}\n")
+            line = 1
+            content = input("Enter content line: ")
+            while content != "stop":
+                file.write(f"Line{line} {content}")
```
Nattalli, line 26: without `Line` word:
`file.write(f"{line} {content}")`

Nattalli: add "\n" in the end of every line

Nattalli: and don't forget to add "\n" for every new appending


should be something like this:
Снимок экрана 2022-10-25 в 18 16 40

Nattalli (changes requested):

Nattalli (changes requested):
Good, one more thing:
In the examples of the task when inserting new text, first comes the empty line, and then the new values

Nattalli (approved):
👍🏻

PR #138, opened 2022-10-09
Solution

LiudmylaKulinchenko:


app/create_file.py
```diff
@@ -1 +1,52 @@
...
+        directories = command[(directory_flag + 1):]
+
+        create_directory(directories)
+
+    elif "-f" in command:
+        file_name = command[-1]
+        add_content(file_name)
+
+
+def create_directory(directory_path: list[str]) -> str:
```
mltdll, line 32: In this function you should use:
- `os.path.join()` instead of `str.join()`;
- `exist_ok=True` flag in `os.makedirs()`;
- `os.path.sep` instead of `"/"`.

app/create_file.py
```diff
@@ -1 +1,52 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory_file() -> None:
```
mltdll, line 6: You can rewrite this function with only two checks, one for -d and one for -f, and with way less code duplication.
Better yet, you can also use `argparse` module here — it would be easier, more readable, and you won't have to reinvent the wheel.

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+    os.makedirs(directory_path)
+
+    return directory_path + "/"
+
+
+def add_content(path: str) -> None:
+    current_time = datetime.strftime(datetime.now(), "%Y-%d-%m %H:%M:%S")
+
+    with open(path, "w") as f:
+        f.write(current_time + "\n")
```
mltdll, line 43: You may use string interpolation instead of concatenation here. Or you can add `"\n"` directly in the `strftime()`.

mltdll (changes requested):

mltdll (approved):
Nice!

PR #140, opened 2022-10-09
Solution

pavlo-kapitanets:


TPolina (changes requested):
Can you put your code in a function or even split it into few functions?

TPolina (approved):

PR #145, opened 2022-10-09
Solution

Qusert17:


app/create_file.py
```diff
@@ -1 +1,51 @@
...
+            index_f = command.index("-f")
+            index_d = command.index("-d")
+            path = "/".join(command[(index_d + 1): index_f])
+            new_directory(path)
+            file_name = "/".join([path, file_name])
+            create_a_file(file_name)
+        create_a_file(file_name)
+        if "-d" in command:
+            index_d = command.index("-d")
+            path = "/".join(command[(index_d + 1):])
```
sbublinskyi, line 46: it is better to use `os.path.join` to make the code more universal for different platforms.

sbublinskyi (changes requested):

app/create_file.py
Qusert17: Done!

sbublinskyi (approved):
nice

PR #147, opened 2022-10-10
create file from terminal

VitaliyBychevoy:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+def create_directories(folders: list) -> None:
+    for i in range(len(folders)):
+        os.mkdir(folders[i])
+        os.chdir(folders[i])
+        print(os.getcwd())
+
+
+def work_with_file(file_name: str) -> None:
+    with open(file_name, "a") as file:
+        time = datetime.now().strftime("%y-%m-%d %H:%M:%S") + "\n"
```
mltdll, line 16: You can add `\n` directly to the string you pass to `.strftime()`.

app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+from datetime import datetime
+import os
+
+import sys
+
+
+def create_directories(folders: list) -> None:
+    for i in range(len(folders)):
```
mltdll, line 8: You can use `os.makedirs()` here. It also has an optional parameter that will be useful here: make sure to check its documentation.

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        os.mkdir(folders[i])
+        os.chdir(folders[i])
+        print(os.getcwd())
```
mltdll, line 11: I don't think this `print()` is necessary here. Same for the other things you print.

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+            file.write(line + "\n")
+            line = input("Enter content line:")
+        file.write("\n")
+
+
+def controller() -> None:
+
+    command_list = sys.argv
+
+    if "-d" in command_list and "-f" not in command_list:
```
mltdll, line 31: You can simplify this function using only two checks. Better yet, you can use `argparse` module, and in this case it will be easier to do this, plus the code would look cleaner.

mltdll (changes requested):

mltdll (approved):

MasterpieceElbow (approved):

PR #148, opened 2022-10-10
Solution

TetyanaPavlyuk:


app/create_file.py
```diff
@@ -1 +1,27 @@
...
+    if "python create_file.py" in line.rstrip():
+        file_data: list[str] = line.rstrip().split()
+        path = ""
+        if "-d" in file_data:
+            no_index: int = file_data.index("-d") + 1
+            while file_data[no_index] != "-f":
+                path = os.path.join(path, file_data[no_index])
+                no_index += 1
+            os.makedirs(path, exist_ok=True)
+        if "-f" in file_data:
```
IvanRamyk, line 15: otherwise, path remains an empty value and line 18 will cause an error

app/create_file.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+for line in sys.stdin:
```
IvanRamyk, line 5: it's easier to use `sys.argv`

IvanRamyk: But you can choose your own approach

IvanRamyk (changes requested):

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+        with open(path, "a") as f:
+            f.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
+            input_data = ""
+            while True:
+                input_data = input("Enter content line: ")
+                if input_data == "stop":
+                    f.write("\n")
+                    break
+                f.write(f"{input_data}\n")
+            break
```
IvanRamyk, line 34: then you don't need these breaks

app/create_file.py
```diff
@@ -1 +1,34 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+for line in sys.stdin:
```
IvanRamyk, line 5: please write a function with all logic
also, think about splitting the whole functionality into several functions that perform different steps of the task

app/create_file.py
```diff
@@ -1,34 +1,43 @@
...
-            f.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
-            input_data = ""
-            while True:
-                input_data = input("Enter content line: ")
-                if input_data == "stop":
-                    f.write("\n")
-                    break
-                f.write(f"{input_data}\n")
-            break
+            raise ValueError
```
DmytroSvirsa, line 40: provide exeption message

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #150, opened 2022-10-11
checklist added

Y-Havryliv:


checklist.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+Bad example:
+
+```python
+if condition:
+    break
+else:
+    print("Bad practice")
+```
```
DmytroSvirsa, line 22: break work inside loops, it may disappoint students

checklist.md
```diff
@@ -0,0 +1,74 @@
...
+with open("file.txt", "a") as f:
+    f.write(f"{i} line\n")
+```
+
+## Clean Code
+
+1. Add comments, prints, and functions to check your solution when you write your code. 
+Don't forget to delete them when you are ready to commit and push your code.
+
+2. Make sure you've tested your code with different flags and it works correctly.
```
DmytroSvirsa, line 74: it`s important, we need to add more text, or give an example

DmytroSvirsa (changes requested):

checklist.md
Y-Havryliv: students use this code snippet several times

checklist.md
Abnormaltype: > it`s important, we need to add more text, or give an example

about what? we have examples in the README.md

Abnormaltype (approved):

DmytroSvirsa (approved):

PR #151, opened 2022-10-11
'Solution'

Ilyakson:


app/create_file.py
```diff
@@ -1 +1,26 @@
...
+
+def create_file() -> None:
+    if "-d" in sys.argv:
+        index = 2
+        while len(sys.argv) != index and sys.argv[index] != "-f":
+            os.mkdir(sys.argv[index])
+            os.chdir(sys.argv[index])
+            index += 1
+    if "-f" in sys.argv:
+        with open(sys.argv[-1], "w") as file:
```
mltdll, line 14: The file should be opened in `append` mode.

mltdll (changes requested):
Overall, the solution seems valid, but it could be simplified using pre-invented wheels such as `argparse` and `os.makerirs()`.

mltdll (approved):

TPolina (approved):
Good job!

PR #152, opened 2022-10-11
Solution

ShevchukVitalii:


app/create_file.py
```diff
@@ -1 +1,30 @@
...
+        return
+    file_path = ""
+    if "-d" in command:
+        file_path = os.path.join(*command[2:-2])
+        if not os.path.exists(file_path):
+            os.makedirs(file_path)
+    file_name = os.path.join(file_path, command[-1])
+
+    with open(file_name, "a") as file:
+        file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
```
mltdll, line 21: You can avoid using f-string if you add `\n` to the end of the string you pass to `.strftime()`.

app/create_file.py
```diff
@@ -1 +1,30 @@
...
+import sys
+from datetime import datetime
+
+
+def create_file() -> None:
+    command = sys.argv
+    if "-f" not in command:
+        file_path = os.path.join(*command[2:])
+        if not os.path.exists(file_path):
+            os.makedirs(file_path)
```
mltdll, line 11: You can remove the check if you use set `exist_ok=True` flag in `os.makedirs()`.

mltdll (changes requested):
The solution seems valid, but it can be cleaner if you use `argparse` module.

mltdll (approved):

app/create_file.py
```diff
@@ -1 +1,29 @@
...
+        file_path = os.path.join(*command[2:-2])
+        if not os.path.exists(file_path):
+            os.makedirs(file_path)
+    file_name = os.path.join(file_path, command[-1])
+
+    with open(file_name, "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+        line_input = input("Enter content line: ")
+        line_number = 1
+        while line_input != "stop":
```
Nattalli, line 23: In the examples of the task when inserting new text, first comes the empty line, and then the new values

Nattalli (changes requested):

Nattalli (approved):
well done!

PR #154, opened 2022-10-11
Solution

mariia-movchan:


app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory(path: list) -> None:
+    os.makedirs("/".join(path))
```
mltdll, line 7: It's better to use `os.path.join()` to provide cross-platform compatibility.
I'd also suggest to set `exist_ok` parameter of `os.makedirs()` to `True`, otherwise it will throw an error if all the directories already exist.

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+    while True:
+        content_line = input("Enter content line: ")
+        if content_line == "stop":
+            break
+        content.append(content_line)
+
+    with open(file_name, "a") as file:
+        file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
+
+        for idx, line in enumerate(content):
```
mltdll, line 21: You can pass an additional `start` argument to `enumerate()` instead of adding 1 to each index.

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+            file.write(f"{idx + 1} {line}\n")
+
+
+def create_file_from_terminal() -> None:
```
mltdll, line 25: This function can be rewritten with only two checks.
But it's even better and much easier to use `argparse` module here. The code will also look cleaner this way.

mltdll (changes requested):

app/create_file.py
mariia-movchan: Thanks, fixed

app/create_file.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory(path: list) -> None:
+    path_ = os.path.join(*path)
```
mltdll, line 7: I'd simply reassign `path` here, but that might be just my preference.

app/create_file.py
mltdll: It's not quite what i had in mind, but I think I can approve. But don't you want to take a look at `argparse`? :)

mariia-movchan: @mltdll I've read about argparse, but if I understood correctly - when argparse is used then both my function should be extended with verification which arguments were passed and are they passed at all?
Hence, there will be extra code.
That's why I consider that it is rationally to left code as is :)

mltdll: It's sufficient to simply create a parser, add two arguments, and use them just as you use the arrays of arguments now. No extra code needed to make the program behave the way it does now.
It is going to be even shorter.

app/create_file.py
mariia-movchan: In my opinion, It's not the best idea to redefine variable path in function.
Let's wait for the mentor?)

mltdll: Sure.

app/create_file.py
mariia-movchan: Done! Thx, it's really useful knowledge  and my code become shorter;)

app/create_file.py
```diff
@@ -1 +1,29 @@
...
+from datetime import datetime
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-f", "--file", help="File name")
+parser.add_argument("-d", "--dir", nargs="+", help="Path")
+command = parser.parse_args()
+
+
+if command.dir:
+    path_ = os.path.join(*command.dir)
```
mltdll, line 12: It seems that you don't have an argument named `path` now. So why don't you use that name? :)

app/create_file.py
```diff
@@ -1 +1,29 @@
...
+if command.file:
+    content = []
+    while True:
+        content_line = input("Enter content line: ")
+        if content_line == "stop":
+            break
+        content.append(content_line)
+
+    with open(command.file, "a") as file:
+        file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
```
mltdll, line 26: You can add `\n` directly to the string in `strftime()`. You can also use formatting options instead, e. g.
```python
file.write(f"{datetime.now():%Y-%m-%d %H:%M:%S\n}")

**Review by mltdll — CHANGES_REQUESTED** [review]

File: app/create_file.py [hunk 7]
```diff
@@ -1 +1,29 @@
...
+
+        for idx, line in enumerate(content, start=1):
```
mltdll, line 28: `index` a short word, you don't lose to much if you don't abbreviate it. Therefore it's probably better to use whole word.

app/create_file.py
mariia-movchan: Really, after all changes, I have no such name )

app/create_file.py
mariia-movchan: I agree

app/create_file.py
mariia-movchan: wow, ok
I improved my code

mltdll (approved):

Nattalli (changes requested):
one comment left but good solution

app/create_file.py
```diff
@@ -1 +1,29 @@
...
+    path = os.path.join(*command.dir)
+    os.makedirs(path, exist_ok=True)
+    os.chdir(path)
+
+
+if command.file:
+    content = []
+    while True:
+        content_line = input("Enter content line: ")
+        if content_line == "stop":
```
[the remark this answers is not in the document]
mariia-movchan: I fixed. By the way, if I add 'file.write("\n")' to the end of context manager I'll get the same result. Or maybe will be better to rewrite code and insert loop WHILE TRUE inside of context manager?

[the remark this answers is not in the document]
Nattalli: in fact, the difference is small and both solutions will be correct

Nattalli (approved):
Good job!

PR #156, opened 2022-10-13
Create file from terminal

LavrykN:


app/create_file.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import argparse
+import os
+from datetime import datetime
+
+parser = argparse.ArgumentParser()
```
IvanRamyk, line 5: let's implement logic inside of the function/functions to be able to reuse them

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #157, opened 2022-10-13
Solution

fsocie7y:


app/create_file.py
```diff
@@ -1 +1,38 @@
...
+def create_directory_and_file() -> None:
+    parse = argparse.ArgumentParser()
+    parse.add_argument("-d", nargs="*")
+    parse.add_argument("-f", nargs=1)
+
+    args = vars(parse.parse_args())
+
+    if args["d"]:
+        make_directory(*args["d"])
+        if args["f"]:
```
mltdll, line 34: This `if` block shouldn't be inside the pevious one. The program should be able to create a file without `-d` argument.

mltdll (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+            add_content_to_file(os.path.join(*args["d"] + args["f"]))
+    else:
```
mltdll, line 36: You can do it without nesting `if` or using `else`. For example,
```python
    if args["d"]:
        make_directory(*args["d"])

    if args["f"]:
        add_content_to_file(os.path.join(*args["d"] or [], *args["f"]))
```
See for yourself if you can improve the code style in the given sample.

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+    os.makedirs(os.path.join(*dir_path), exist_ok=True)
+
+
+def add_content_to_file(*path) -> None:
+    current_time = datetime.datetime.now().strftime("%Y-%d-%m %H:%M:%S")
+
+    with open(os.path.join(*path), "a") as file:
+        file.write(f"{current_time}\n")
+        num_line = 0
+        content = input("Enter content line:")
```
mltdll, line 16: By the way, it's better to add a space at the end of the string you pass to `input()`.

mltdll (changes requested):

app/create_file.py
fsocie7y: sorry, i don't have other idea

mltdll (approved):

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+        while content.lower() != "stop":
+            num_line += 1
+            file.write(f"{num_line} {content}\n")
+            content = input("Enter content line:")
+            if content == "stop":
```
MasterpieceElbow, line 21: Why do you need this if you had the oposit condition in while? You can do it right after the loop.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #158, opened 2022-10-16
Solution of py-create-file-from-terminal

YLobunets:


app/create_file.py
```diff
@@ -1 +1,56 @@
...
+            for line in lines:
+                file_obj.write(line + "\n")
+
+
+if __name__ == "__main__":
+    program_argv = sys.argv
+    # # program_argv = ["231231231231.py","-d","aaa","bbb"]
+    # program_argv = ["231231231231.py","-d","aaa","bbb","-f", "ccc.txt"]
+    # print(program_argv)
+    create_file(program_argv)
```
DmytroSvirsa, line 56: make reading from terminal a part of function

DmytroSvirsa (changes requested):
thik ho to reduse code size

app/create_file.py
YLobunets: I made command line reading as a part of function

DmytroSvirsa (approved):

PR #166, opened 2022-11-01
Solution create file from terminal

anton-chumak-main:


app/create_file.py
```diff
@@ -1 +1,55 @@
-# write your code here
+import os
+from datetime import datetime
+
+
+def flags_choice() -> None:
```
IvanRamyk, line 5: use `sys.args` to parse flags entered from the terminal

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #169, opened 2022-11-04
Solution

GeorgePavlej:


app/create_file.py
```diff
@@ -1 +1,30 @@
-# write your code here
+from datetime import datetime
+import sys
+import os
+
+command = sys.argv
+join_folders = ""
+
+if "-d" in command and len(sys.argv) > 2:
```
IvanRamyk, line 8: let's create a function/functions to have the opportunity to reuse this code

IvanRamyk (changes requested):

DmytroSvirsa (approved):

PR #170, opened 2022-11-05
Solution

Terrrya:


app/create_file.py
```diff
@@ -1 +1,55 @@
...
+
+if dir_names and file_name:
+    file_create(file_name, path_to_file=make_dir(dir_names))
+    quit()
+
+if dir_names:
+    make_dir(dir_names)
+
+if file_name:
+    file_create(file_name)
```
DmytroSvirsa, line 55: provide function for this code

DmytroSvirsa (changes requested):

app/create_file.py
Terrrya: Done

DmytroSvirsa (approved):

PR #172, opened 2022-11-07
'Sol_#1'

Jumanje82:


app/create_file.py
```diff
@@ -1 +1,23 @@
...
+        s.write("\n")
+
+
+if "-d" in sys.argv and "-f" in sys.argv:
+    os.makedirs("/".join(sys.argv[2:-2]))
+    create_file("/".join(sys.argv[2:-2]) + "/" + sys.argv[-1])
+elif "-f" in sys.argv:
+    create_file(sys.argv[-1])
+elif "-d" in sys.argv:
+    os.makedirs("/".join(sys.argv[-2:]))
```
DmytroSvirsa, line 23: make it as function and you can do it without duplicating "-d" in sys.argv and "-f" in sys.argv only for 2 statements

DmytroSvirsa (changes requested):

app/create_file.py
```diff
@@ -1 +1,28 @@
...
+            if new == "stop":
+                break
+            s.write(new + "\n")
+        s.write("\n")
+
+
+def create_from_terminal() -> None:
+    if "-d" in sys.argv and "-f" in sys.argv:
+        os.makedirs("/".join(sys.argv[2:-2]))
+        os.chdir(str("/".join(sys.argv[2:-2])))
```
Serhii-Leonenko, line 20: use `os.path.join()` and don't duplicate it, you can create one variable for it.

app/create_file.py
```diff
@@ -1 +1,28 @@
...
+        create_file()
+    elif "-f" in sys.argv:
+        create_file()
+    elif "-d" in sys.argv:
+        os.makedirs("/".join(sys.argv[-2:]))
```
Serhii-Leonenko, line 25: the same about `os.path.join()`

Serhii-Leonenko (changes requested):

OleksandrPozden (approved):
Nice!

PR #173, opened 2022-11-09
Solution

yessloss:


DmytroSvirsa (changes requested):
avoid code duplication - create function for repeating actions

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+
+def create_file(new_file: str) -> None:
+    count = 0
+    new_line_file = ""
+    result_name_file = new_file.split()
+    if "python" == result_name_file[0]:
+        for index in range(len(result_name_file)):
+            if result_name_file[index] == "-d":
+                for index_f in range(len(result_name_file)):
+                    if "-f" in result_name_file:
```
DmytroSvirsa, line 13: here

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+                        if result_name_file[index_f] == "-f":
```
DmytroSvirsa, line 14: here

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+                            directories = result_name_file[index + 1: index_f]
+                            for directory in directories:
+                                if os.path.exists(directory) is False:
+                                    os.mkdir(directory)
+                    if "-f" not in result_name_file:
```
DmytroSvirsa, line 19: here

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+                        directories = result_name_file[index + 1:]
+                        for directory in directories:
+                            if os.path.exists(directory) is False:
+                                os.mkdir(directory)
+            if result_name_file[index] == "-f":
```
DmytroSvirsa, line 24: here

DmytroSvirsa (changes requested):
also you have 3 leves of for loop
create functions to optimize your code and have the only 1 check for tag -f

DmytroSvirsa (changes requested):
ok, that what i want you to do:
1) create function for create folder
2) create function for write data in file
3) in create_file function:
- first check the -d tag and call function for create folder
- than check the -f tag and call function for write data in file

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+
+
+def create_file(new_file: str) -> None:
+    result_name_file = new_file.split()
+    if "python" == result_name_file[0]:
+        for index in range(len(result_name_file)):
+            if result_name_file[index] == "-d":
+                create_dir(result_name_file, index)
+            if result_name_file[index] == "-f":
+                write_data_in_file(result_name_file, index)
```
DmytroSvirsa, line 12: you can use `"-d" in result_name_file` to avoid this for loop, also what will happen if I have 4 "-f" tags it will create a file for each -f?

DmytroSvirsa (changes requested):

app/create_file.py
yessloss: but I need index

app/create_file.py
```diff
@@ -5,11 +5,12 @@
...
-        for index in range(len(result_name_file)):
-            if result_name_file[index] == "-d":
-                create_dir(result_name_file, index)
-            if result_name_file[index] == "-f":
-                write_data_in_file(result_name_file, index)
+        if "-d" in result_name_file:
+            index = result_name_file.index("-d")
+            create_dir(result_name_file, index)
+        if "-f" in result_name_file:
+            index2 = result_name_file.index("-f")
```
Serhii-Leonenko, line 12: `index2` is a bad name for variable

app/create_file.py
```diff
@@ -5,11 +5,12 @@
 def create_file(new_file: str) -> None:
```
Serhii-Leonenko, line 5: You should work with `sys.argv` instead of  that your fuction takes str parametr.

app/create_file.py
```diff
@@ -5,11 +5,12 @@
...
     result_name_file = new_file.split()
     if "python" == result_name_file[0]:
```
Serhii-Leonenko, line 7: You don't need this check

app/create_file.py
```diff
@@ -5,11 +5,12 @@
...
-        for index in range(len(result_name_file)):
-            if result_name_file[index] == "-d":
-                create_dir(result_name_file, index)
-            if result_name_file[index] == "-f":
-                write_data_in_file(result_name_file, index)
+        if "-d" in result_name_file:
```
Serhii-Leonenko, line 8: I recommend you to add one more check for your function, and make it first:
`if "-d" in command and "-f" in command:`
       `create directories....`
       `write data to file...`
 It allows you to remove unnecessary checks  like in  `create_dir` and unnecessary fuction like `for_duplicates`

app/create_file.py
```diff
@@ -19,19 +20,19 @@ def for_duplicate(directories: list[str]) -> None:
...
         for_duplicate(directories)
 
 
-def write_data_in_file(result_name_file: list[str], index: int) -> None:
+def write_data_in_file(result_name_file: list[str], index2: int) -> None:
     new_line_file = ""
     count = 0
-    name_file = "".join(result_name_file[index + 1:])
+    name_file = "".join(result_name_file[index2 + 1:])
     with open(name_file, "w") as source_file:
```
Serhii-Leonenko, line 36: You should use `"a"` parametr instead of `"w"`

app/create_file.py
```diff
@@ -19,19 +20,19 @@ def for_duplicate(directories: list[str]) -> None:
...
         time_now = datetime.datetime.now()
         source_file.write(time_now.strftime("%Y-%m-%d %H:%M:%S"))
```
Serhii-Leonenko, line 38: You can use `while True` loop, because you have check `stop` inside the loop. And you should have `\n` after time row.

app/create_file.py
```diff
@@ -5,11 +5,12 @@
...
-                write_data_in_file(result_name_file, index)
+        if "-d" in result_name_file:
+            index = result_name_file.index("-d")
+            create_dir(result_name_file, index)
+        if "-f" in result_name_file:
+            index2 = result_name_file.index("-f")
+            write_data_in_file(result_name_file, index2)
 
 
 def for_duplicate(directories: list[str]) -> None:
```
Serhii-Leonenko, line 16: Are you sure that this function will create correct directories ?) If you have `-d dir1 dir2` it means that you should create `dir1` inside current directory  and `dir2` inside `dir1`

Serhii-Leonenko (changes requested):

app/create_file.py
yessloss: is it necessary?

app/create_file.py
```diff
@@ -1 +1,55 @@
-# write your code here
+import os
+import sys
+import datetime
+
+
+def create_file(name_file: str) -> None:
+    #new_file = sys.argv
```
Abnormaltype, line 7: remove unnecessary comments

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+        while "stop" != new_line_file:
+            new_line_file = input("Enter content line: ")
+            count += 1
+            if "stop" == new_line_file:
+                break
+            source_file.write(f"{count} "
+                              f"Line{count} {new_line_file}\n")
+
+
+create_file("python create_file.py -d dir1 dir2 -f file.txt")
```
Abnormaltype, line 55: all your tests put in the if name == main construction

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #174, opened 2022-11-09
Create file

bythewaters:


DmytroSvirsa (changes requested):
too many checks of "-f" tag create aditional function, and make it with only one "-f" tag check

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+            name_file += file_name
+
+    line_char = 1
+
+    while True:
+        input_data = str((input("Enter content line: ")))
+        if input_data == "stop":
+            break
+        data_file += str(line_char) + " " + input_data + "\n"
+        line_char += 1
```
DmytroSvirsa, line 55: you can put the input statement inside the with statement and wright it directly to the file it will work much better with memory

DmytroSvirsa (changes requested):

app/create_file.py
bythewaters: Oh, it's really work faster, but why? This is due to the hash function? 
just moving to the next topic)

app/create_file.py
```diff
@@ -39,26 +39,24 @@ def crate_path(terminal_command: list) -> str:
...
+            data_file = ""
+            while True:
+                input_data = str((input("Enter content line: ")))
+                if input_data == "stop":
+                    break
+                data_file += str(line_char) + " " + input_data + "\n"
+                line_char += 1
+
             file.write(datetime.now().strftime("%Y-%m-%d, %H:%M:%S") + "\n")
             file.write(data_file + "\n")
```
DmytroSvirsa, line 61: you can make it even faster if you will write each input separately in file inside the while loop

DmytroSvirsa (changes requested):

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #176, opened 2022-11-18
'Solution'

anastasiia-hnusova:


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+import argparse
+import os
+from datetime import datetime as dt
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument("-d", type=str,
+                    default=[],
+                    nargs="*",
+                    help="all items after this flag are parts of the path")
```
OleksandrPozden, line 10: for better readability, it is better to style this code like this
```
parser.add_argument(
         "-d",
         type=str,
         ....
)
```

OleksandrPozden (changes requested):
good, but one minor mistake

YuriiKindrat (approved):

Abnormaltype (approved):

PR #177, opened 2022-11-22
Solution

the-andrew-007:


app/create_file.py
```diff
@@ -1 +1,42 @@
-# write your code here
+import os
+import sys
+from datetime import datetime as dt
+
+if __name__ == "__main__":
```
Serhii-Leonenko, line 5: You should create a function, and then you can test it in `if __name__ == "__main__":`

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+
+    dir_path = ""
+    if dirs_parts:
+        dir_path = os.path.join(*dirs_parts)
+        if not os.path.exists(dir_path):
+            os.makedirs(dir_path)
+
+    if filename:
+        file_content = str(dt.now().strftime("%Y-%m-%d, %H:%M:%S")) + "\n"
+        while (line := input("Enter content line: ")) != "s":
```
Serhii-Leonenko, line 33: image

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+            file_content += line + "\n"
+        file_content = file_content.rstrip()
+        with open(os.path.join(dir_path, filename), "a") as users_file:
+            users_file.seek(0, os.SEEK_END)
+            file_is_empty = not bool(users_file.tell())
+            users_file.seek(0)
+            if not file_is_empty:
+                users_file.write("\n")
+            users_file.write(file_content)
```
Serhii-Leonenko, line 42: When you append new data to the file which exists, there should be an empty line between data, I didn't see it:
image

Serhii-Leonenko (changes requested):

OleksandrPozden (approved):
Good to go!

PR #178, opened 2022-11-27
Solution

Ihor-Ivashov:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+        with open(os.path.join(full_path, file_name), "a") as file:
+            file.write(dt.strftime(dt.now(), "%Y-%m-%d %H:%M:%S") + "\n")
+            count_line = 0
+            while True:
+                line = input("Enter content line: ")
+                if line == "stop":
+                    file.write("\n")
+                    break
+                count_line += 1
+                file.write(f"{count_line} {line}\n")
```
OleksandrPozden, line 17: Do not overwhelm the context manager. Split the input logic and the writing to the file logic

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+            mode = "create_dir"
+            continue
+        if arg == "-f":
+            mode = "create_file"
+            continue
+
+        if mode == "create_dir":
+            full_path = os.path.join(full_path, arg)
+        elif mode == "create_file":
+            file_ls.append(arg)
```
OleksandrPozden, line 35: simplify this

OleksandrPozden (changes requested):

app/create_file.py
Ihor-Ivashov: It was universal solution for all variants of arguments. For example create_file.py -f file1.txt file2.txt -d dir1 dir2

YuriiKindrat (approved):

PR #183, opened 2022-12-02
Solution

MykytaMoshchenko:


app/create_file.py
```diff
@@ -1 +1,38 @@
...
+from sys import argv
+import os
+from datetime import datetime
+
+
+def create_file() -> None:
+    command = argv
+    flag_d = "-d"
+    flag_f = "-f"
+
```
OleksandrPozden, line 10: these variables better initialize outside the function as const

OleksandrPozden (changes requested):

app/create_file.py
MykytaMoshchenko: Thank you, changed.

MykytaMoshchenko: @OleksandrPozden 

Please, have a look at the updated solution.
Thank you in advance!

app/create_file.py
```diff
@@ -1 +1,38 @@
-# write your code here
+from sys import argv
+import os
+from datetime import datetime
+
+flag_d = "-d"
```
Abnormaltype, line 5: constant should be written in the uppercase

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+
+        while True:
+            content = input("Enter content line: ")
+            if content == "stop":
+                break
+            created_file.write(f"{line} {content}\n")
+            line += 1
+
+
+create_file()
```
Abnormaltype, line 38: don't run your program

Abnormaltype (changes requested):

Abnormaltype (approved):

PR #184, opened 2022-12-02
Solution

eduard-olentsevych:


app/create_file.py
```diff
@@ -1 +1,46 @@
...
+        current_date_time = datetime.datetime.now()
+        timestamp = current_date_time.strftime("%Y-%m-%d %H:%M:%S")
+        new_file.write(timestamp + "\n")
+        line_number = 1
+        while True:
+            content = input("Enter content line: ") + "\n"
+            if content == "stop\n":
+                break
+            new_file.write(f"{line_number} {content}")
+            line_number += 1
```
OleksandrPozden, line 24: "\n" has to be rewritten.
```suggestion
            content = input("Enter content line: ")
            if content == "stop":
                break
            new_file.write(f"{line_number} {content}\n")
            line_number += 1
```

OleksandrPozden (changes requested):

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+    if "-f" in command and "-d" not in command:
+        create_file(command[-1])
+
+    if "-f" in command and "-d" in command:
+        new_directories = command[command.index("-d") + 1: command.index("-f")]
+        path_to_new_directory = create_directory(new_directories)
+        path_to_new_file = \
+            os.path.join(path_to_new_directory, f"{command[-1]}")
+        create_file(path_to_new_file)
```
Serhii-Leonenko, line 42: If you put this condition before both previous, you can remove unnecessary `"-f" not in command` and `"-d" not in command`

Serhii-Leonenko (changes requested):

app/create_file.py
```diff
@@ -28,19 +28,19 @@ def main() -> None:
...
         new_directories = command[command.index("-d") + 1: command.index("-f")]
         path_to_new_directory = create_directory(new_directories)
         path_to_new_file = \
             os.path.join(path_to_new_directory, f"{command[-1]}")
         create_file(path_to_new_file)
 
+    elif "-d" in command:
+        create_directory(command[command.index("-d") + 1:])
+
+    elif "-f" in command:
```
OleksandrPozden, line 41: use if instead of elif

OleksandrPozden (changes requested):

app/create_file.py
eduard-olentsevych: @OleksandrPozden 
If I change "elif" to "if", then, as I understand, all the conditions in all three "if" are met in the case of "python create_file.py -d dir1 dir2 -f file.txt", and the directories up to "-f" and "file.txt" (as directories) will be created. Would it be acceptable to leave "elif"?

PR #191, opened 2022-12-13
Solution

Sebshe:


app/create_file.py
```diff
@@ -1 +1,51 @@
...
+        if "-f" not in terminal_input:
+            make_dirs(
+                home_path,
+                terminal_input[terminal_input.index("-d") + 1:]
+            )
+        else:
+            make_dirs(
+                home_path,
+                terminal_input[
+                    terminal_input.index("-d") + 1: terminal_input.index("-f")
```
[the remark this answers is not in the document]
IvanRamyk: same for terminal_input.index("-f")

app/create_file.py
```diff
@@ -1 +1,51 @@
-# write your code here
+import os
+import datetime
+import sys
+
+
+def make_dirs(abs_path: str, dirs_to_create: list[str]) -> None:
```
IvanRamyk, line 6: it seems that this logic could be simplified

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+    for directory in dirs_to_create:
+        abs_path = os.path.join(abs_path, directory)
+    os.makedirs(abs_path)
```
Grayder0152, line 9:
```suggestion
    os.makedirs(abs_path, exists_ok=True)
```

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+                break
+            line_number += 1
+            new_file.write(f"{line_number} {content}\n")
+
+
+def create_file() -> None:
+    home_path = os.getcwd()
+    terminal_input = sys.argv
+    f_index = terminal_input.index("-f") if "-f" in terminal_input else False
+    d_index = terminal_input.index("-d") if "-d" in terminal_input else False
```
Grayder0152, line 33: Use method `find` instead of `index`

app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+import os
+import datetime
+import sys
```
Grayder0152, line 3: Sort imports

Grayder0152 (changes requested):

app/create_file.py
Sebshe: terminal_input is a list created by sys.arvg. There is no "find" method for lists

Vasyl-Smutok (approved):
Good Job 👍

PR #196, opened 2022-12-22
'Solution'

Oleksiy-Liubchenko:


app/create_file.py
```diff
@@ -1 +1,49 @@
...
+    timestamp = current_time.strftime("%Y-%m-%d %H:%M:%S")
+
+    with open(file_name, "a") as write_file:
+        write_file.write("\n" + timestamp + "\n")
+        while True:
+            output = input("Enter content line:") + "\n"
+            if output != "stop" + "\n":
+                write_file.write(output)
+            else:
+                break
```
Serhii-Leonenko, line 49: Make your script executable, add` if __name__ == "__main__": `

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        create_new_file(command)
+
+
+def create_new_directory(path_list: list) -> str:
+    path_ = ""
+    for folder in path_list:
+        path_ += folder + "/"
+        if os.path.exists(path_) is not True:
+            os.mkdir(path_)
+    return path_
```
Serhii-Leonenko, line 31: Read the checklist. Use `os.path.join()`

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+    else:
+        file_name = command.split()[-1]
+    current_time = datetime.now()
+    timestamp = current_time.strftime("%Y-%m-%d %H:%M:%S")
+
+    with open(file_name, "a") as write_file:
+        write_file.write("\n" + timestamp + "\n")
+        while True:
+            output = input("Enter content line:") + "\n"
+            if output != "stop" + "\n":
```
Serhii-Leonenko, line 46: why "stop" +"\n" ?

Serhii-Leonenko: and change this condition, remove redundant else

Serhii-Leonenko (changes requested):

app/create_file.py
Oleksiy-Liubchenko: Hi, did not understand why we need this construction? I googled it, they say if __name__ == "__main__" add when we want to launch our program as script, not as module. But previously I launched my program only with "run" in pycharm.

Am I correct and this construction need for what I wrote? And how we can run our program as script? 
P.S. maybe I did it before, but did not know that it called "run as script"))

Oleksiy-Liubchenko: @Serhii-Leonenko

app/create_file.py
Oleksiy-Liubchenko: without + "\n" in console when i paste "stop" it is not working and append it like a string in file

Oleksiy-Liubchenko:
I changed code but without if name = main. How to do it correctly? 

if name == "main":
    create_new_file()

and in parentheses what i need to set?

app/create_file.py
YuriiKindrat: u can remove "\n" from ur input function and add it to ur write method
also no need ot use else here u can just make one check for stop and break if input == stop

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,50 @@
...
+        write_file.write("\n" + timestamp + "\n")
+        while True:
+            output = input("Enter content line:")
+            if output == "stop":
+                break
+            write_file.write(output + "\n")
+
+
+# if "__name__" == "__main__":
+#     create_new_file()
```
OleksandrPozden, line 50:
```suggestion

if __name__ == "__main__":
     create_new_file()
```

OleksandrPozden (changes requested):

app/create_file.py
Oleksiy-Liubchenko: I did it, but could you please, answer to my questions above about this ?

OleksandrPozden (approved):

PR #207, opened 2023-01-06
Create file from terminal

MarynaProkhorenko:


app/create_file.py
```diff
@@ -1 +1,35 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+list_of_input_data = sys.argv
```
olena-hul, line 6: Separate your code into functions to make it more modular

app/create_file.py
```diff
@@ -1 +1,35 @@
...
+if "-d" in list_of_input_data:
+    path_to_file = os.join.path(
+        list_of_input_data[0],
```
olena-hul, line 9: For what do you need first argument? It is the name of the script

olena-hul (changes requested):

app/create_file.py
```diff
@@ -1 +1,31 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_file_and_input_data(path_to_new_file: str) -> None:
+    with open(path_to_new_file, "a") as data_to_write:
+        date_now = datetime.now()
+        data_to_write.write(date_now.strftime("%m/%d/%Y, %H:%M:%S") + "\n")
```
katevasilenko, line 9: Date time format is not the same as required

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+            data_to_write.write(line + "\n")
+            line = input("Enter content line:")
+
+
+def create_file_from_terminal() -> None:
+    list_of_input_data = sys.argv
+    if "-d" in list_of_input_data:
+        path_to_file = os.join.path(
+            list_of_input_data[2],
+            list_of_input_data[3]
```
katevasilenko, line 21: Does that make sense to split those above and named those variables?

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+        )
+        os.makedirs(path_to_file)
+
+        if "-f" in list_of_input_data:
+            create_file_and_input_data(
+                os.join.path(path_to_file, list_of_input_data[-1])
```
katevasilenko, line 27: For this and 31 line `list_of_input_data` too

katevasilenko (changes requested):

app/create_file.py
MarynaProkhorenko: Не розумію в чому тут проблема. Поясніть, будь ласка

katevasilenko: Return to the previous code and just declare some variables like `directory_path`, `file_name` whatever for your `list_of_input_data[-1] `, `list_of_input_data[1] ` etc.
It makes your code more clear to read and quick understanding. Does that make sense?

katevasilenko (changes requested):
I left comment under yours

olena-hul:
Good job!

olena-hul (approved):

PR #212, opened 2023-01-12
solution

xpanzhar:


app/create_directory.py
```diff
@@ -0,0 +1,30 @@
+import sys
+import os
+
+
+def create_directory() -> str:
+    """Function parse terminal input, creates directory
+     and  return file name"""
+
+    input_list = sys.argv
+    file_name = "file.txt"
```
katevasilenko, line 10: You should not create a file, if you have only `-d` flag. But your code creates `file.txt ` by default wherever if `-f` flag was given or not.

app/create_directory.py
```diff
@@ -0,0 +1,30 @@
...
+
+    elif "-d" in input_list:
+        path = os.path.join(
+            *input_list[input_list.index("-d") + 1:]
+        )
+        os.makedirs(path)
+        os.chdir(path)
+
+    elif "-f" in input_list:
+        file_name = input_list[-1]
```
katevasilenko, line 28: Those `elif` statements are repeat code from the `if` statement

app/create_file.py
```diff
@@ -1 +1,18 @@
-# write your code here
+from datetime import datetime
+from create_directory import create_directory
+
+
+def create_file() -> None:
+    with open(create_directory(), "a") as new_file:
```
katevasilenko, line 6: Here should be full path to file but your function `create_directory ` returns only file name

katevasilenko (changes requested):

Vasyl-Smutok (approved):
Well done!

PR #213, opened 2023-01-12
Solution

nazar-dmytrenko:


app/create_file.py
```diff
@@ -1 +1,45 @@
...
+
+
+def main() -> None:
+    args = sys.argv[1:]
+    file_path = []
+    if "-d" in args and "-f" in args:
+
+        if args[0] == "-d":
+            file_path = [arg for arg in args[1:-2]]
+            file_path.append(args[1])
```
andrewbelo, line 27: Цей рядок повторюється

andrewbelo (changes requested):

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+        else:
+            create_directory(args[args.index("d") + 1:])
+            write_file(f"{os.path.join(*args[args.index('-d') + 1:])}"
+                       f"/{args[1]}")
+
+    elif "-d" in args:
+        create_directory(args[1::])
+
+    elif "-f" in args:
+        write_file(args[-1])
```
OleksandrPozden, line 37: why not check if there is `'-f'` in `args` and then take `args.index('-f')+1` argument and manipulate with it? The same for `'-f'`. Hence you have only 2 if's

OleksandrPozden (changes requested):
Everything is right expect of if-statements

OleksandrPozden (approved):

PR #218, opened 2023-01-21
Solution

aarrtemm:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    if "-d" in sys.argv:
+        index = sys.argv.index("-d")
+        dir_path = "/".join(sys.argv[index + 1:])
+        file_path = f"{dir_path}/{file_name}"
+    else:
+        file_path = file_name
+    if os.path.exists(file_path):
+        with open(file_path, "a") as file:
+            timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            file.write(f"timestamp\n")
```
ArsenPidhoretskyi: You do not need f-string here.
f-string used when we need to pass some value to string for example:
```
a = 1
f'print {a}' # print 1
```

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    file_name = sys.argv[index + 1]
+    if "-d" in sys.argv:
+        index = sys.argv.index("-d")
+        dir_path = "/".join(sys.argv[index + 1:])
+        file_path = f"{dir_path}/{file_name}"
+    else:
+        file_path = file_name
+    if os.path.exists(file_path):
+        with open(file_path, "a") as file:
+            timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
ArsenPidhoretskyi, line 22: Move format to constant variable

app/create_file.py
```diff
@@ -1 +1,43 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+if "-d" in sys.argv:
+    index = sys.argv.index("-d")
+    dir_path = "/".join(sys.argv[index + 1:])
+    os.makedirs(dir_path, exist_ok=True)
```
ArsenPidhoretskyi, line 9: Check what will be created in next case:

ArsenPidhoretskyi: You need to check -f flag also. It can be before -d and also after -d

app/create_file.py
ArsenPidhoretskyi: Btw, as I see you forgot to pass timestamp as value, not as string

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+            timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            file.write(f"timestamp\n")
+
+            i = 1
+            while True:
+                content = input("Enter content line: ")
+                if content == "stop":
+                    break
+                file.write(f"{i} {content}\n")
+                i += 1
```
ArsenPidhoretskyi, line 43: Code duplicate, only changes in mode

ArsenPidhoretskyi (changes requested):
1. Rewrite to function
2. Check code duplicates

app/create_file.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+dir_path = ""
+file_name = ""
+
+for i in range(len(sys.argv)):
```
Vasyl-Smutok, line 9: `i` bad variable name

app/create_file.py
```diff
@@ -1 +1,28 @@
...
+    elif sys.argv[i] == "-f":
+        file_name = sys.argv[i + 1]
+
+    file_path = os.path.join(dir_path, file_name)
+    mode = "a" if os.path.isfile(file_path) else "w"
+
+    with open(file_path, mode) as file:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(timestamp + "\n")
+    i = 1
```
Vasyl-Smutok, line 22: `i` bad variable name

Vasyl-Smutok (changes requested):

app/create_file.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+dir_path = ""
+file_name = ""
+
+for string in range(len(sys.argv)):
```
olena-hul, line 9: Bad name for index

olena-hul (changes requested):

app/create_file.py
```diff
@@ -3,26 +3,40 @@
...
-file_name = ""
-
-for string in range(len(sys.argv)):
-    if sys.argv[string] == "-d":
-        dir_path = "/".join(sys.argv[string + 1:])
-        os.makedirs(dir_path)
-    elif sys.argv[string] == "-f":
-        file_name = sys.argv[string + 1]
-
+def create_file(dir_path: str, file_name: str) -> callable:
```
olena-hul, line 6: Return annotation is wrong

app/create_file.py
```diff
@@ -3,26 +3,40 @@
...
+    os.makedirs(dir_path, exist_ok=True)
     file_path = os.path.join(dir_path, file_name)
     mode = "a" if os.path.isfile(file_path) else "w"
-
     with open(file_path, mode) as file:
```
olena-hul, line 10: Just pass "a" instead of mode. It will be enough

olena-hul (changes requested):

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+        new_file.write(f"{counter} {content} \n")
+        counter += 1
+
+
+def main(args: list) -> None:
+    dir_path = ""
+    file_name = ""
+    for index in range(len(args)):
+        if args[index] == "-d":
+            dir_path = "/".join(args[index + 1:])
```
ArsenPidhoretskyi, line 31: Use os.path.join
Because in different os it will works in another way

ArsenPidhoretskyi: BTW, it will works bad with:
python main.py -d dir1/dir2 -f index.txt
you dir_path=dir1/dir2/-f/index.txt/

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+        elif args[index] == "-f":
+            file_name = args[index + 1]
+    if not dir_path or not file_name:
```
ArsenPidhoretskyi, line 34: Due to requirements:

> If only -d flag passed, means all items after this flag are parts of the path.

> If only -f flag passed, means first item is the file name.

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+
+
+def create_file(dir_path: str, file_name: str) -> callable:
+    os.makedirs(dir_path, exist_ok=True)
+    file_path = os.path.join(dir_path, file_name)
+    mode = "a" if os.path.isfile(file_path) else "w"
+    with open(file_path, mode) as file:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(timestamp + "\n")
+    return file
```
ArsenPidhoretskyi, line 13: I think your file will be closed. No need to return it, better to open it again

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+
+
+def write_content(new_file: object) -> None:
```
ArsenPidhoretskyi, line 16: Better to accept file_name

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+        content = input("Enter content line: ")
+        if content == "stop":
+            break
+        new_file.write(f"{counter} {content} \n")
+        counter += 1
+
+
+def main(args: list) -> None:
+    dir_path = ""
+    file_name = ""
```
ArsenPidhoretskyi, line 28: Why do you need to declare these arguments?

ArsenPidhoretskyi (changes requested):

katevasilenko (approved):
Looks good to me

PR #225, opened 2023-01-26
Solution

6aPa6aH:


app/create_file.py
```diff
@@ -1 +1,45 @@
...
+            i += 1
+        file.write("\n")
+
+
+def create_dir_and_file(command: list) -> None:
+    file_name = command[-1]
+    path_dir = ""
+    if "-d" in command and "-f" in command:
+        for i in command[2:-2]:
+            path_dir += f"{i}/"
```
olena-hul, line 33: Use os.path.join to make the directory name consistent for all OS. And are you sure that directory path will be exatly from indexes 2 to -2? What if I pass -f flag first, and -d flag second?

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        create_dir(path_dir)
+        create_file(path_dir + file_name)
```
olena-hul, line 35: os.path.join

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    elif "-f" in command:
+        create_file(file_name)
+    elif "-d" in command:
+        for i in command[2:]:
+            path_dir += f"{i}/"
```
olena-hul, line 40: os.path.join

app/create_file.py
```diff
@@ -1 +1,45 @@
-# write your code here
+import os
+import datetime
+import sys
+
+
+def create_dir(path_dir: str) -> None:
+    if not os.path.exists(path_dir):
+        os.makedirs(os.path.dirname(path_dir))
```
olena-hul, line 8:
```suggestion
    os.makedirs(os.path.dirname(path_dir), exist_ok=True)
```

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+
+
+def create_file(file_name: str) -> None:
+    file_attribute = "w"
+    if os.path.exists(file_name):
+        file_attribute = "a"
```
olena-hul, line 14: You can use only "a" mode

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    with open(f"{file_name}", f"{file_attribute}") as file:
```
olena-hul, line 15: No need to use f-strings here

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        i = 1
```
olena-hul, line 16: i - bad variable name

olena-hul (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+                break
+            file.write(f"{line_number} {new_line} \n")
+            line_number += 1
+        file.write("\n")
+
+
+def create_dir_and_file(command: list) -> None:
+    path_dir = os.path.abspath(os.curdir)
+    if "-d" in command and "-f" in command and\
+            command.index("-d") < command.index("-f"):
```
Vasyl-Smutok, line 27:
```
if ("-d" in command and "-f" in command and 
    command.index("-d") < command.index("-f")):
```

better to use `()` than `/`

Vasyl-Smutok (approved):
Good Job 👍

app/create_file.py
6aPa6aH: I tried that, but flake8 issues error: "app/create_file.py:26:45: W504 line break after binary operator"

PR #226, opened 2023-01-30
Solution

alina-boichenko:


app/create_file.py
```diff
@@ -1 +1,42 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory() -> None:
+    input_command = sys.argv
```
Abnormaltype, line 7: You can make it as CONSTANT, b/c u don't change this variable

Abnormaltype (approved):
Good job!
Only one small suggestion

PR #233, opened 2023-02-09
'Solution'

Paul-Maslov:


app/create_file.py
```diff
@@ -1 +1,24 @@
...
+for i in range(1, amount_of_argv):
+    if sys.argv[i] == "-f":
+        now = datetime.datetime.now()
+        with open(f"{name_dir}/{str(sys.argv[i + 1])}", "w") as f:
+            f.write(f"{now.date()} {now.hour}:{now.minute}:{now.second}\n")
+            for line in sys.stdin:
+                if "stop" == line.rstrip():
+                    break
+                print(f"Enter content line: {line}")
+                f.write(line)
```
Bohdan-Salamakha, line 24: even if u create a shell script u should create a program startpoint, in python it is:
```
if __name__ == "__main__":
    some_func(...)
```

app/create_file.py
```diff
@@ -1 +1,24 @@
-# write your code here
+import sys
+import datetime
+import os
+
+
+amount_of_argv = len(sys.argv)
+name_dir = sys.path[0]
+if sys.argv[1] == "-d":
```
Bohdan-Salamakha, line 8: put sys.argv into separated variable

app/create_file.py
```diff
@@ -1 +1,24 @@
...
+    for item in range(2, amount_of_argv):
+        if sys.argv[item] == "-f":
+            break
+        name_dir = name_dir + "/" + str(sys.argv[item])
+        os.mkdir(name_dir)
+
+for i in range(1, amount_of_argv):
```
Bohdan-Salamakha, line 15: divide ur program  into several functions

Bohdan-Salamakha (changes requested):
v1

app/create_file.py
Paul-Maslov: Done

app/create_file.py
Paul-Maslov: Done

app/create_file.py
Paul-Maslov: done

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+        for line in sys.stdin:
+            if "stop" != line.rstrip():
+                print(f"Enter content line: {line}")
+                f.write(line)
+                continue
+            break
+
+
+if __name__ == "__main__":
+    command_list = sys.argv
```
Bohdan-Salamakha, line 33: usually we call one func or classmethod at the startpoint so u just need to create one more func named like main()
```
def main():
...
if __name__ == "__main__":
    main()
```

Bohdan-Salamakha (changes requested):
v2

app/create_file.py
Paul-Maslov: Done

Bohdan-Salamakha (approved):

PR #236, opened 2023-02-13
Solution

SlavikSherbak:


app/create_file.py
```diff
@@ -1 +1,51 @@
...
+        filename: str,
+        regime: str,
+        context: str
+) -> None:
+    with open(filename, regime) as file:
+        file.write(context)
+
+
+def create_directories(dir_names: list) -> None:
+    os.makedirs(os.path.join(*dir_names))
```
Grayder0152, line 30: use `exist_ok` parameter

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+
+elif "-f" in command and "-d" in command:
+    create_directories(
+        command[(command.index("-d") + 1):command.index("-f")]
+    )
+    command.remove("-f")
+    command.remove("-d")
+    create_file(
+        os.path.join(*command), "w", create_content()
+    )
```
Grayder0152, line 51: Add this code to function

Grayder0152 (changes requested):

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+    with open(filename, regime) as file:
+        file.write(context)
+
+
+def create_directories(dir_names: list) -> None:
+    os.makedirs(os.path.join(*dir_names), exist_ok=True)
+
+
+def analysis_command() -> None:
+    if "-f" in command and "-d" not in command:
```
ArsenPidhoretskyi, line 34: Not so good to work with global
Need to accept command parameter

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+        file_name = command[1]
+        if os.path.exists(file_name):
+            create_file(file_name, "a", f"\n{create_content()}")
+        else:
+            create_file(file_name, "w", create_content())
```
ArsenPidhoretskyi, line 39: Better to move this check to create_file function

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+
+    elif "-f" not in command and "-d" in command:
+        create_directories(command[1:])
+
+    elif "-f" in command and "-d" in command:
+        create_directories(
+            command[(command.index("-d") + 1):command.index("-f")]
```
ArsenPidhoretskyi, line 46: It is not true)
I want -f flag first and then -d

ArsenPidhoretskyi: Then i fill have something like this)
file1.txt/dir1/dir2

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,54 @@
-# write your code here
+import sys
+import os
+import datetime
```
Grayder0152, line 3: Pls order imports

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+        create_file(command_text[1], create_content())
+
+    elif "-f" not in command_text and "-d" in command_text:
+        create_directories(command_text[1:])
+
+    elif "-f" in command_text and "-d" in command_text:
+        command_text.remove("-f")
+        command_text.remove("-d")
+        create_directories(
+            command_text[0:2]
```
Grayder0152, line 47: Why `[0:2]`? Parts of the path can be more or less than 2

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+def create_file(
+        filename: str,
+        context: str
+) -> None:
+    if os.path.exists(filename):
+        with open(filename, "a") as file:
+            file.write(f"\n{context}")
+    else:
+        with open(filename, "a") as file:
+            file.write(context)
```
Grayder0152, line 29: Pls remove dublicate from your code

Grayder0152 (changes requested):

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+
+    elif "-f" not in command_text and "-d" in command_text:
+        create_directories(command_text[1:])
+
+    elif "-f" in command_text and "-d" in command_text:
+        command_text.remove("-f")
+        command_text.remove("-d")
+        create_directories(
+            command_text[0:-1]
+        )
```
ArsenPidhoretskyi, line 48: Not works when I want -f first then -d

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+    elif "-f" not in command_text and "-d" in command_text:
+        create_directories(command_text[1:])
+
+    elif "-f" in command_text and "-d" in command_text:
+        command_text.remove("-f")
+        command_text.remove("-d")
+        print(command_text)
+        if command_text[0] == "file.txt":
+            command_text.reverse()
+            command_text[0], command_text[1] = command_text[1], command_text[0]
```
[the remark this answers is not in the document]
SlavikSherbak: This code makes some changes to the list with the command. My code works correctly only if the command looks like this (-d dir1 dir2 -f file.txt), in case there is -f in the command first, that is, the command will look like this (-f file.txt -d dir1 dir2) I created this code that will convert a list of str into the sequence I need so that everything works correctly.

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+        create_file(command_text[1], create_content())
+
+    elif "-f" not in command_text and "-d" in command_text:
+        create_directories(command_text[1:])
+
+    elif "-f" in command_text and "-d" in command_text:
+        command_text.remove("-f")
+        command_text.remove("-d")
+        print(command_text)
+        if command_text[0] == "file.txt":
```
ArsenPidhoretskyi, line 47: I do not need the file name equals to file.txt
I want main.txt for example
You code will broke

app/create_file.py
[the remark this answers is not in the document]
ArsenPidhoretskyi: It does not work correctly also)
You code must work with all cases:
1. file name not only file.txt
2. len dirs is not only two it can be one or 3 or 10 for example

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done)

PR #242, opened 2023-02-16
Solution

olenaliuby:


app/create_file.py
```diff
@@ -1 +1,59 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_path(directories: list) -> str:
+    current_directory = os.path.dirname(__file__)
+    path = os.path.join(current_directory, *directories)
+    os.makedirs(path)
```
ArsenPidhoretskyi, line 9: add exist_ok=True

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+                break
+            line_number += 1
+            file.writelines(f"{line_number} {file_line}\n")
+
+
+def create_file_from_terminal() -> None:
+    terminal_input = sys.argv
+    current_directory = os.path.dirname(__file__)
+
+    if terminal_input[1] == "-f":
```
ArsenPidhoretskyi, line 34: if terminal input have length equal to 1

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        file_name = terminal_input[2]
+        file_path = os.path.join(current_directory, file_name)
```
ArsenPidhoretskyi, line 36: ???? If -d flag is present than your code will not work

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        create_file_with_content_from_input(file_path)
+
+    try:
+
+        if terminal_input[1] == "-d" and "-f" not in terminal_input:
+            directories_list = terminal_input[2:]
+            create_path(directories_list)
+
+        if "-d" in terminal_input and "-f" in terminal_input:
+            directories_list = terminal_input[2:terminal_input.index("-f")]
```
ArsenPidhoretskyi, line 46: Not work if -f flag will be first

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+            file_name = terminal_input[-1]
```
ArsenPidhoretskyi, line 47: -f flag can be first also

ArsenPidhoretskyi (changes requested):

app/create_file.py
olenaliuby: Thank you for highlighting those bugs. I hope that my changes made the app better.

Grayder0152 (approved):

PR #243, opened 2023-02-17
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/create_file.py
+++ b/app/create_file.py
@@ -1,1 +1,64 @@
-# write your code here
+from datetime import datetime
+import os.path
+import sys
+
+command_line = sys.argv
+
+path_to_file = ""
+file_name = ""
+
+
+def create_path(directories: list) -> str:
+    path = os.path.join(*directories)
+    return path
+
+
+def print_to_file(file_name: str, flag: bool) -> None:
+    content = datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n"
+    line_number = 1
+    while True:
+        if flag is False:
+            print_line = input(f"Enter content line: Line{line_number} ")
+            if print_line == "stop":
+                content += "\n"
+                break
+            content += f"Line{line_number} {print_line}\n"
+        else:
+            print_line = input(
+                f"Enter content line: Another line{line_number} ")
+            if print_line == "stop":
+                content += "\n"
+                break
+            content += f"Another line{line_number} {print_line}\n"
+
+        line_number += 1
+
+    with open(file_name, "w" if flag is False else "a") as file_out:
+        file_out.write(content)
+
+
+if "-d" in command_line and "-f" not in command_line:
+    directories = command_line[command_line.index("-d") + 1:]
+    path_to_file = create_path(directories)
+
+    if not os.path.exists(path_to_file):
+        os.makedirs(path_to_file)
+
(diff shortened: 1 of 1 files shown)


app/create_file.py
```diff
@@ -1 +1,64 @@
...
+    path = os.path.join(*directories)
+    return path
+
+
+def print_to_file(file_name: str, flag: bool) -> None:
+    content = datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n"
+    line_number = 1
+    while True:
+        if flag is False:
+            print_line = input(f"Enter content line: Line{line_number} ")
```
AndereLion, line 21: Your result of writing must be like this "Enter content line:"  +   content  .     "Line1 content" is "content"

AndereLion (changes requested):

app/create_file.py
SafonovVladimir: Fixed

AndereLion (approved):
Gj

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+def create_directory(file_name: str = None) -> None:
+    if file_name:
+        path = create_path(sys.argv[2:-2])
+    else:
+        path = create_path(sys.argv[2:])
+
+    try:
+        os.makedirs(path)
+    except FileExistsError:
+        return
```
YuriiKindrat, line 34: no need to use blank return here, u can use exist_ok arg for makedirs func

YuriiKindrat (changes requested):

app/create_file.py
SafonovVladimir: I don't know about this. Thank's

kuznetsb (approved):
+

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+    line_number = 1
+
+    with open(sys.argv[-1], "a") as file_out:
+        file_out.seek(0, os.SEEK_END)
+        data = datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n"
+
+        if file_out.tell() == 0:
+            content = data
+        else:
+            content = "\n" + data
```
YuriiKindrat, line 21: no need to make this check here, u can just add blank line before break keyword

YuriiKindrat (changes requested):

app/create_file.py
SafonovVladimir: Done

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+                break
+            file_out.write(f"{line_number} {content}\n")
+            line_number += 1
+
+
+def create_directory(file_name: str = None) -> None:
+    if file_name:
+        path = create_path(sys.argv[2:-2])
+    else:
+        path = create_path(sys.argv[2:])
```
YuriiKindrat, line 30: `path = sys.argv[2:] if file_name else sys.argv[2:-2]`
`path = create_path(path)`
or
`path = sys.argv[2:]`
`if filename:`
`path = sys.argv[2:-2]`
`path = create_path(path)`
what do u think abouth that?

YuriiKindrat (approved):

app/create_file.py
SafonovVladimir: Yeh, it's easier for reading.
Thank's.
I change my code

PR #244, opened 2023-02-17
1 commit

AndereLion:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    directory = ""
+    flag = False
+    for word in dir_path:
+        if word == "-f":
+            create_file(directory + filename)
+            break
+        if flag:
+            directory += f"{word}/"
+            os.mkdir(os.path.join(directory))
+        if word == "-d":
```
SafonovVladimir, line 20: It's better to make directories like list, and append to that list.
After os.path.join(*directories)

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+def create_file(file_name: str) -> None:
+    with open(os.path.join(file_name), "a") as file:
+        command = str(datetime.datetime.today())[:-7]
+        file.write(command + "\n")
+        while True:
+            command = input("Enter content line: ")
+            if command == "stop":
+                break
+            file.write(command + "\n")
+        file.write("\n")
```
SafonovVladimir, line 33: Where is "Another line1" and line counter?

app/create_file.py
```diff
@@ -1 +1,43 @@
-# write your code here
+import os
+import sys
+import datetime
+
+
+# os.makedirs(os.path.join("dir1/dir2/"))
```
SafonovVladimir, line 6: remove it

SafonovVladimir (changes requested):

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+        if flag:
+            directory.append(word)
+            os.mkdir(os.path.join(*directory))
+        if word == "-d":
+            flag = True
+
+
+def create_file(file_name: str) -> None:
+    with open(os.path.join(file_name), "a") as file:
+        command = str(datetime.datetime.today())[:-7]
```
SafonovVladimir, line 24: Better to use .strftime

SafonovVladimir (changes requested):

bihunva (approved):
GJ

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+    for word in dir_path:
+        if word == "-f":
+            directory.append(filename)
+            create_file(os.path.join(*directory))
+            break
+        if flag:
+            directory.append(word)
+            os.mkdir(os.path.join(*directory))
+        if word == "-d":
+            flag = True
```
YuriiKindrat, line 19: what if we swap the flags ?

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+
+if len(sys.argv) > 1 and "-d" in sys.argv and "-f" in sys.argv:
+    create_directory(" ".join(sys.argv), sys.argv[-1])
+
+elif len(sys.argv) > 1 and "-d" in sys.argv:
+    create_directory(" ".join(sys.argv))
+
+elif len(sys.argv) > 1 and "-f" in sys.argv:
+    create_file(sys.argv[-1])
```
YuriiKindrat, line 41: make ur code DRY no need to make this `" ".join(sys.argv), sys.argv[-1]` twice - make a variables for this

YuriiKindrat (changes requested):

app/create_file.py
AndereLion: Fixed.

app/create_file.py
AndereLion: What do you mean?  =)

app/create_file.py
```diff
@@ -1 +1,43 @@
-# write your code here
+import os
+import sys
+import datetime
+
+
+def create_directory(dir_path: str, filename: str = None) -> None:
+    dir_path = dir_path.split(" ")
```
YuriiKindrat, line 7: no need to use space as arg in split, it uses space by default

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    for word in dir_path:
+        if word == "-f":
+            directory.append(filename)
+            create_file(os.path.join(*directory))
+            break
+        if d_in_directory:
+            directory.append(word)
+            os.mkdir(os.path.join(*directory))
+        if word == "-d":
+            d_in_directory = True
```
YuriiKindrat, line 19: what if command will be like this `python create_file.py -f file.txt -d dir1 dir2`

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+
+
+def create_file(file_name: str) -> None:
+    with open(os.path.join(file_name), "a") as file:
+        command = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(command + "\n")
+        while True:
+            command = input("Enter content line: ")
+            if command == "stop":
+                break
```
YuriiKindrat, line 29: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

YuriiKindrat (changes requested):

app/create_file.py
AndereLion: Fixed this bug )

app/create_file.py
AndereLion: Added

app/create_file.py
AndereLion: Done

YuriiKindrat (approved):

PR #247, opened 2023-02-21
solution

avkpol:


app/create_file.py
```diff
@@ -1 +1,39 @@
-# write your code here
+import os
+from datetime import datetime as d
```
Grayder0152, line 2: do not use alias for datetime, is it not clear

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+create_file("python create_file.py -d dir1 dir2")
+create_file("python create_file.py -f file.txt")
+create_file("python create_file.py -d dir1 dir2 -f file.txt")
+
+"""
+
+
+def create_file(input_string: str) -> None:
+    _, __, flag, *destination_path = input_string.split()
+    dir_path = [s for s in destination_path if "-f" not in s]
```
Grayder0152, line 14: what if i want to create diractory `dir-fox` ? Lets use `s!='-f'` instead of `"-f" not in s`

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+"""
+create_file("python create_file.py -d dir1 dir2")
+create_file("python create_file.py -f file.txt")
+create_file("python create_file.py -d dir1 dir2 -f file.txt")
+
+"""
+
+
+def create_file(input_string: str) -> None:
+    _, __, flag, *destination_path = input_string.split()
```
Grayder0152, line 13: slice result of splite method to avoid use `_ , __`

Grayder0152 (changes requested):

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+    elif flag == "-d":
+        os.makedirs(dirpath, exist_ok=True)
+    elif flag == "-f":
+        with open(output_file, "a") as target_file:
+            write_content(target_file)
+
+
+def write_content(target_file: callable) -> None:
+    ts = datetime.timestamp(datetime.now())
+    ts = datetime.fromtimestamp(int(ts))
```
ArsenPidhoretskyi, line 30: Why so hard???
datetime.now() does not work?
You do: datetime.now() -> to timestamp -> to datetime.now()

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+"""
+create_file("python create_file.py -d dir1 dir2")
+create_file("python create_file.py -f file.txt")
+create_file("python create_file.py -d dir1 dir2 -f file.txt")
+
+"""
+
+
+def create_file(input_string: str) -> None:
+    lg, f_name, flag, *destination_path = input_string.split()
```
ArsenPidhoretskyi, line 13: what mean lg

ArsenPidhoretskyi (changes requested):
Your code does not work with this command:

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+    elif flag == "-f":
+        if "-d" in path_:
+            if str_.index("-f") < str_.index("-d"):
+                dr, filename = path_[2:], path_[0]
+                return dr, filename
+        filename = path_[-1]
+        return [], filename
+
+
+def create_file(path_: str) -> None:
```
ArsenPidhoretskyi, line 31: rename path_ to command

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+from pathlib import Path
+"""
+create_file("python create_file.py -d dir1 dir2")
+create_file("python create_file.py -f file.txt")
+create_file("python create_file.py -d dir1 dir2 -f file.txt")
+create_file("python create_file.py -f file.txt -d dir1 dir2")
+"""
+
+
+def check_str(input_string: str) -> tuple:
```
ArsenPidhoretskyi, line 12: rename check_str to parse_command

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+    str_ = input_string.split()
+    unused1, unused2, flag, *path_ = str_
```
ArsenPidhoretskyi, line 14: Use _ instead of unused
It is so bad to create not used variables

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+"""
+create_file("python create_file.py -d dir1 dir2")
+create_file("python create_file.py -f file.txt")
+create_file("python create_file.py -d dir1 dir2 -f file.txt")
+create_file("python create_file.py -f file.txt -d dir1 dir2")
+"""
+
+
+def check_str(input_string: str) -> tuple:
+    str_ = input_string.split()
```
ArsenPidhoretskyi, line 13: str_ is so bad name
It is not clear

app/create_file.py
ArsenPidhoretskyi, line 12: It can return None also

ArsenPidhoretskyi: Also tuple of what?

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+    unused1, unused2, flag, *path_ = str_
+    if flag == "-d":
+        if "-f" in path_:
+            if str_.index("-d") < str_.index("-f"):
+                dr, filename = path_[0:-2], path_[-1]
+                return dr, filename
+        dr = path_[0:]
+        return dr, []
```
ArsenPidhoretskyi, line 21: Just do:
return path_, ""

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+create_file("python create_file.py -f file.txt -d dir1 dir2")
+"""
+
+
+def check_str(input_string: str) -> tuple:
+    str_ = input_string.split()
+    unused1, unused2, flag, *path_ = str_
+    if flag == "-d":
+        if "-f" in path_:
+            if str_.index("-d") < str_.index("-f"):
```
ArsenPidhoretskyi, line 17: BTW, due to your code in this case it is always true
because -d have index 2 and f always > 3

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+                dr, filename = path_[0:-2], path_[-1]
+                return dr, filename
+        dr = path_[0:]
+        return dr, []
+    elif flag == "-f":
+        if "-d" in path_:
+            if str_.index("-f") < str_.index("-d"):
```
ArsenPidhoretskyi, line 24: The same with always true

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+                dr, filename = path_[2:], path_[0]
+                return dr, filename
+        filename = path_[-1]
+        return [], filename
+
+
+def create_file(path_: str) -> None:
+    dr, filename = check_str(path_)
+    if not dr:
```
ArsenPidhoretskyi, line 33: simplify this code:
Only two if statement:
1. if dr -> create directories
2. if filename -> create content

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+        dirpath = os.path.join(*dr)
+        os.makedirs(dirpath, exist_ok=True)
+    elif dr and filename:
+        pat = dr + [filename]
+        dirpath = os.path.join(*pat)
+        Path(dirpath).parent.mkdir(exist_ok=True, parents=True)
+        write_content(dirpath)
+
+
+def write_content(_file: str) -> None:
```
ArsenPidhoretskyi, line 45: rename _file to file_path

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+        current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        target_file.write(current_time + "\n")
+        number = 0
+        while True:
+            new_line = input("Enter content line:")
+            if new_line == "stop":
+                target_file.write("\n")
+                break
+            number += 1
+            target_file.write("%d  %s" % (number, new_line + "\n"))
```
ArsenPidhoretskyi, line 56: Use f-string it looks better

ArsenPidhoretskyi (changes requested):
Before re-request review check that all variables names are clear:
https://geo-python.github.io/site/notebooks/L1/gcp-1-variable-naming.html
https://www.rithmschool.com/blog/good-ideas-for-better-variable-names
https://inventwithpython.com/beyond/chapter4.html

app/create_file.py
avkpol: Screenshot 2023-02-24 at 11 21 30
who should I follow?! )))

app/create_file.py
avkpol: but dr can be not empty and in such case if use Path(dirpath).parent.mkdir(exist_ok=True, parents=True) and "dir" will be the last element in the path,  it will create "dir" as a filename and write content to it.

avkpol: but there are in fact 3 cases `['dir1', 'dir2'] []`,`[] file.txt`, `['dir1', 'dir2'] file.txt` How to handle it with two conditions?

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+from typing import Any
+"""
+create_file("python create_file.py -d dir1 dir2")
+create_file("python create_file.py -f file.txt")
+create_file("python create_file.py -d dir1 dir2 -f file.txt")
+create_file("python create_file.py -f file.txt -d dir1 dir2")
+"""
+
+
+def parse_command(input_string: str) -> tuple[list, Any] | None:
```
ArsenPidhoretskyi, line 13:
```suggestion
def parse_command(input_string: str) -> tuple[list, str] | None:
```

app/create_file.py
ArsenPidhoretskyi:
```
directories_path = os.path.join(*directories)

if directories_path:
    os.makedirs(directories_path, exist_ok=True)
    
    
if file_name:
    file_path = os.path.join(directories_path, file_name)
    # and then write inside file 

```

ArsenPidhoretskyi: As for there are two general case:
is directories exists
and is file_name exists

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+    split_str = input_string.split()
+    _, _, flag, *path_ = split_str
```
ArsenPidhoretskyi, line 15:
```suggestion
    _, _, flag, *path_ = input_string.split()
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #248, opened 2023-02-21
Solution

WDemidenko:


app/create_file.py
```diff
@@ -1 +1,23 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+commands = [sys.argv[i] for i in range(1, len(sys.argv))]
```
NEkropo1, line 5: it's ok, but better to slice string, instead calling sys func couple of times and make array out of that, seems like heavily overweighted, you could do just commands = sys.argv[1:]

app/create_file.py
```diff
@@ -1 +1,23 @@
...
+commands = " ".join(commands).split("-f")
+command_path = commands.pop(0).strip("-d").split()
```
NEkropo1, line 7: and if you just sliced input in way I mentioned above these commands should've been more easy, like split("-f:)[0]
sure you got what I mean

app/create_file.py
```diff
@@ -1 +1,23 @@
...
+if command_path:
+    path = [folder for folder in command_path]
+    path = os.path.join(*path)
+    if not os.path.exists(path):
+        os.makedirs(path)
+if commands:
+    with open(commands[0], "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d, %H:%M:%S") + "\n")
```
NEkropo1, line 15: didn't notice this before, make more namespaces, really, a **honking** great idea, so code would be much more readable, for example you could make output_file = commands[0] before this block and it would be more readable -> refactorable just in case
Just compare `if commands` to `if output_file
`

NEkropo1 (changes requested):

NEkropo1 (approved):
Well, it's up to you, I still don't like your error handling, you know what I mean, but it's working as expected and quite readable now, gj!

YuriiKindrat (changes requested):
Separate your code into functions to make it more modular

YuriiKindrat (approved):

PR #249, opened 2023-02-21
Solution

MaybeTI:


SafonovVladimir (approved):
GJ

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+            if line_text == "stop":
+                break
+            file.write(f"1 {line_text}\n")
+
+
+def create_file_in_directory() -> None:
+    os.makedirs("/".join(command[2: command.index("-f")]))
+
+    with open(f"{'/'.join(command[2: command.index('-f')])}/{command[-1]}",
+              "w") as file:
```
YuriiKindrat, line 27: it is realy hard to read this, simplify this part, f.e. make variables for this parts

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+        direc_file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
+        line = 1
+        while True:
+            line_text = input("Enter content line: ")
+
+            if line_text == "stop":
+                direc_file.write("\n")
+                break
+            direc_file.write(f"{line} {line_text}\n")
+            line += 1
```
YuriiKindrat, line 36: why do u need this part ? u can use create_file func here, make ur code DRY

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+
+def create_file() -> None:
+    with open(command[-1], "w") as file:
+        file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
+        while True:
+            line_text = input("Enter content line: ")
+            if line_text == "stop":
+                break
+            file.write(f"1 {line_text}\n")
```
YuriiKindrat, line 20: what about lines here? also add blank line after stop word)

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+
+def create_file_in_directory() -> None:
+    os.makedirs("/".join(command[2: command.index("-f")]))
+    direc_file = f"{'/'.join(command[2: command.index('-f')])}/{command[-1]}"
```
YuriiKindrat, line 25: u can make variable from this part `'/'.join(command[2: command.index('-f')]` an reuse it in makedirs and direc_file, no need to make join for 2 times

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+import os
+import sys
+from datetime import datetime
+
+
+command = sys.argv
+
+
+def make_dir() -> None:
+    os.makedirs("/".join(command[2:]))
```
YuriiKindrat, line 10: no need to create separate method for another method

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+    direc_file = "/".join(command[2: command.index("-f")])
+    os.makedirs(direc_file)
+    create_file(f"{direc_file}/")
+
+
+if "-d" in command and "-f" not in command:
+    make_dir()
+
+elif "-f" in command and "-d" not in command:
+    create_file()
```
YuriiKindrat, line 37: no need to use "not in command" condition here, u can just move ur last condition to the top

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+elif "-f" in command and "-d" in command:
```
YuriiKindrat, line 39: no need to use not in command condition here, move ur last condition to the top

app/create_file.py
MaybeTI: I can't understand what should I do

YuriiKindrat: @MaybeTI why do u need a separate method to call another method inside it ? u can just call method makedirs where do u need

YuriiKindrat (changes requested):
image
also, fix this everywhere

PR #251, opened 2023-02-22
Solution

NEkropo1:


DavidGlusch (approved):

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+if __name__ == "__main__":
+    if "-d" in sys.argv and "-f" in sys.argv:
+        dir_path = os.path.join(*sys.argv[2:-1])
+        file_name = sys.argv[-1]
+        os.makedirs(dir_path, exist_ok=True)
+        file_path = os.path.join(dir_path, file_name)
+        create_file(file_path)
+    elif "-d" in sys.argv:
+        dir_path = os.path.join(*sys.argv[2:])
+        os.makedirs(dir_path, exist_ok=True)
```
YuriiKindrat, line 31: u can write separate func for this

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+        with open(file_path, "a") as f:
+            f.write("\n")
+    else:
+        with open(file_path, "w") as f:
+            f.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+
+    while True:
+        line = input("Enter content line : ")
+        if line == "stop":
+            break
```
YuriiKindrat, line 17: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+        with open(file_path, "a") as f:
+            f.write(f"{line}\n")
```
YuriiKindrat, line 19: why do u need to use with open for 3 times? optimize that

YuriiKindrat (changes requested):

app/create_file.py
NEkropo1: Where? I see no new lines.
image
image

NEkropo1: image

NEkropo1: If you meant that I should provide lines with /n, like in this screen:
image
My code provides it after taking all input, to every item added:
    `f.write(f"{line}\n")`

YuriiKindrat (approved):

PR #255, opened 2023-02-22
create-file-from-terminal-Solution

MilManIT:


app/create_file.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+if "-d" in sys.argv:
```
olena-hul, line 5: Move the code into function or functions

app/create_file.py
```diff
@@ -1 +1,20 @@
...
+    dir_path = os.path.join(*sys.argv[sys.argv.index("-d") + 1:])
+    os.makedirs(dir_path, exist_ok=True)
+else:
+    dir_path = "."
+
+if "-f" in sys.argv:
+    file_path = os.path.join(dir_path, sys.argv[sys.argv.index("-f") + 1])
+    with open(file_path, "a" if os.path.exists(file_path) else "w") as f:
```
olena-hul, line 13: You can use a mode always

app/create_file.py
```diff
@@ -1 +1,20 @@
...
+        if f.tell() == 0:
```
olena-hul, line 14: Why do you need this?

olena-hul (changes requested):

app/create_file.py
MilManIT: Check if the '-f' flag is present in the command-line arguments.
Get the file name by looking for the argument immediately after the '-f' flag.
Join the directory path and the file name to get the full file path.
Open the file in 'append' mode if it already exists, or 'write' mode if it doesn't.
If the file is empty, write the current timestamp at the beginning.

Bohdan-Salamakha (changes requested):
don't work on the master branch - bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

Bohdan-Salamakha (changes requested):
don't work on the master branch - it's a bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

PR #256, opened 2023-02-23
Solution

ZabFTFT:


amber-marichi (changes requested):
your code works but there are 2 remarks:
1) why do you have 2 same files in your solution ? :)
2) after writing content in file there are 2 blank lines left after the last line with text. Im not sure this is how it was supposed to be according to the task. Maybe you could check and solve  it?

ZabFTFT:
> your code works but there are 2 remarks:
> 
> 1. why do you have 2 same files in your solution ? :)
> 2. after writing content in file there are 2 blank lines left after the last line with text. Im not sure this is how it was supposed to be according to the task. Maybe you could check and solve  it?

Thanks for suggestions, I will check it and fix

amber-marichi (approved):

amber-marichi (approved):

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+    if first_flag == "-d":
+        path_directories = os.path.join(os.getcwd(), *content)
+        if not os.path.exists(path_directories):
+            os.makedirs(path_directories)
+    elif first_flag == "-f":
+        path = os.path.join(os.getcwd(), *content)
+
+    with open(path, "a") as file_out:
+        if file_out.tell() != 0:
+            file_out.write("\n")
```
YuriiKindrat, line 21: no need to make this check, u can just add blank line before break work, but it is up to u

YuriiKindrat (changes requested):
break ur code to a several function, it is hard to read it like that, also to make it more modular

app/create_file.py
```diff
@@ -1 +1,69 @@
...
+        if args.f and args.d:
+            creates_directories()
+            creates_file(path)
+        elif args.f:
+            creates_file(path)
+        elif args.d:
+            creates_directories()
+    except Exception as e:
+        print(f"{type(e)} was raised: \n{e}")
+        raise
```
YuriiKindrat, line 69: why do u need this blank raise?

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #259, opened 2023-02-23
Solution

kuznetsb:


app/create_file.py
```diff
@@ -1 +1,36 @@
...
+        if user_input == "stop":
+            break
+        user_data.append(f"{len(user_data)} {user_input}\n")
+
+    with open(filename, "a+") as file:
+        file.seek(0)
+        data = file.read()
+        if data:
+            file.writelines(["\n"] + user_data)
+        else:
```
SergiiMachulin, line 31: Maybe you could re-write without else condition here as better practice

SergiiMachulin (approved):
LGTM!

app/create_file.py
```diff
@@ -1 +1,36 @@
...
+import sys
+import os
+from datetime import datetime
+
+
+def main(commands: list) -> None:
+    if "-d" in commands and "-f" not in commands:
+        os.makedirs(os.path.join(*commands[1:]), exist_ok=True)
+    elif "-f" in commands and "-d" not in commands:
+        make_file(commands[-1])
```
YuriiKindrat, line 10: no need to write not in commands condittion here, u can just replace last condtion to the top

app/create_file.py
```diff
@@ -1 +1,36 @@
...
+        os.makedirs(full_path, exist_ok=True)
+        make_file(os.path.join(full_path, file_name))
+
+
+def make_file(filename: str) -> None:
+    user_data = [f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n"]
+    while True:
+        user_input = input("Enter content line: ")
+        if user_input == "stop":
+            break
```
YuriiKindrat, line 23: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,36 @@
...
+        user_data.append(f"{len(user_data)} {user_input}\n")
+
+    with open(filename, "a+") as file:
+        file.seek(0)
+        data = file.read()
+        if data:
+            file.writelines(["\n"] + user_data)
+        else:
+            file.writelines(user_data)
```
YuriiKindrat, line 32: so there won't need to make this check

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #260, opened 2023-02-23
Solution

yura-hudzovskyi:


Morty67 (approved):

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+
+def create_path(direction: list) -> None:
+    if direction:
+        direction = os.path.join(*direction)
+        if not os.path.exists(direction) and len(direction) > 0:
+            os.makedirs(direction)
+            os.chdir(direction)
+            print(f"Path: '{direction}' has been successfully created")
+            return
+        os.chdir(direction)
```
YuriiKindrat, line 18: no need to use blank return here, u are using os.chdir in both cases

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+                file.write("\n")
+                print("Data recording is complete!")
+                break
+            file.write(f"{line_number} {content}\n")
+            line_number += 1
+
+
+if __name__ == "__main__":
+    create_path(path)
+    create_file(commands)
```
YuriiKindrat, line 37: why do u pass argument in create_file as commands, but use it inside ur function as file_name? are u sure that file_name is valid name for arg?

YuriiKindrat (changes requested):

wQuelS (approved):
GJ

PR #263, opened 2023-02-24
'Solution'

RuslanLopatin:


app/create_file.py
```diff
@@ -1 +1,42 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+def create_file() -> None:
+    if "-d" in sys.argv:
```
oksana-feshchenko, line 7: when you make this check, it includes situation when you have -f flag , so it  creates even  directory named "-f"

oksana-feshchenko (changes requested):

PR #266, opened 2023-02-24
'Solution'

Morty67:


viktoria-rybenchuk (approved):

inrock1 (approved):

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+from datetime import datetime
+import sys
+import os
+
+
+def create_derectory() -> None:
+    current = os.getcwd()
+    command = sys.argv
+    command_f = "-f"
+    command_d = "-d"
```
YuriiKindrat, line 10: why do u need to create this variable here, if u'll rewrite it at the next line ?

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+    if command_d in command:
+        for line in range(2, len(command)):
+            if command[line] == command_f:
+                break
+            current = os.path.join(current, command[line])
+        os.mkdir(current)
```
YuriiKindrat, line 16: try to optimize ur code using makedirs metod

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+
+
+def create_file(file_name: str) -> None:
+    with open(file_name, "w") as new_file:
+        new_file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+        line_counter = 1
+        while True:
+            asker = input("Enter content line: ")
+            if asker == "stop":
+                break
```
YuriiKindrat, line 26: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+            new_file.write(f"{line_counter} {asker}")
+            line_counter += 1
+
+
+if __name__ == "__main__":
+    create_file("hello.txt")
```
YuriiKindrat, line 32: so it is always be an hello.txt ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+    if "-d" in command:
+        for line in range(2, len(command)):
+            if command[line] == "-f":
+                break
+            current = os.path.join(os.getcwd(), command[line])
+            os.makedirs(current, exist_ok=True)
+
+
+def create_file() -> None:
+    file_name = input("Enter a name of file")
```
YuriiKindrat, line 17: read a readme.md file again, it should take a name fo file from a command, also it should create a file inside a dirs if the -d flag exist

YuriiKindrat (changes requested):

app/create_file.py
Morty67: > read a readme.md file again, it should take a name fo file from a command, also it should create a file inside a dirs if the -d flag exist

sry, I'm sick right now and my brain doesn't work very well
i have problem with all code, or only file_name variable?

Serhii-Leonenko (changes requested):
Will your script work if you run this command in the terminal `python create_file.py -d dir1 dir2 -f file.txt`?
I think not, so you should make it works with commands like this from the terminal. Read the task again.

YuriiKindrat (approved):

PR #270, opened 2023-02-24
Solution

ada-krav:


me1nyk (approved):
Good job)

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+def create_directory(path: str) -> None:
+    if not os.path.exists(path):
+        os.makedirs(path)
+
+
+def create_file(path: str = None) -> None:
+    if path is not None:
+        file_name = f"{path}\\{args[-1]}"
+    else:
+        file_name = f"{args[-1]}"
```
YuriiKindrat, line 18: image

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+import os
+import sys
+
+
+args = sys.argv[1:]
+
+
+def create_directory(path: str) -> None:
+    if not os.path.exists(path):
+        os.makedirs(path)
```
YuriiKindrat, line 11: no need to create a new func for using another func, u can use exist_ok param for makedirs func

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+if "-d" in args and "-f" not in args:
+    path = os.path.join(*args[1:])
+    create_directory(path)
+
+elif "-d" not in args and "-f" in args:
+    create_file()
+
+elif "-d" in args and "-f" in args:
+    path = os.path.join(*args[:-2])
```
YuriiKindrat, line 38: why do u need "not in args" condition, move ur last condition to the top

YuriiKindrat (approved):

PR #273, opened 2023-02-25
Solution

Yevheniy-Lototskiy:


app/create_file.py
```diff
@@ -1 +1,61 @@
...
+
+
+if "-d" in sys.argv and "-f" not in sys.argv:
+    create_directory()
+
+if "-f" in sys.argv and "-d" not in sys.argv:
+    create_file()
+
+if "-d" in sys.argv and "-f" not in sys.argv:
+    create_directory_and_file()
```
NickSavchyn, line 61: no newline at the end of file

NickSavchyn (changes requested):

app/create_file.py
YuriiKindrat, line 61: no need to use "not" condition here, use elif instead

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+            line += 1
+        f.write("\n")
+
+
+def create_directory_and_file() -> None:
+    directory = [folder for folder in sys.argv[2: sys.argv.index("-f")]]
+
+    create_directory(directory)
+
+    directory = "/".join(directory)
```
YuriiKindrat, line 47: use os.path.join here

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+def create_directory(directory: list = None) -> None:
+    if not directory:
+        directory = [folder for folder in sys.argv[2:]]
+
+    path = directory[0]
+    if len(directory[1:]) > 1:
+        for folder in directory[1:]:
+            os.mkdir(path)
+            path += "/" + folder
+    os.mkdir(path)
```
YuriiKindrat, line 15: u can use makedirs func to simplify ur code

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+def create_file(directory: str = None) -> None:
+    if directory:
+        path = directory
+    else:
+        path = sys.argv[2]
+
+    if os.path.exists(path):
+        mode = "a"
+    else:
+        mode = "w"
```
YuriiKindrat, line 27: u can only use "a" mode here, it will also create a file if it doesn't exist

YuriiKindrat (changes requested):

PR #274, opened 2023-02-25
Solution

oksanaaam:


app/create_file.py
```diff
@@ -1 +1,39 @@
...
+                line = 1
+                while True:
+                    content = input("Enter content line: ")
+                    if content == "stop":
+                        f.write("\n")
+                        break
+                    f.write(f"{line} {content}\n")
+                    line += 1
+        else:
+            time_now = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
ostapT, line 26: you have this code duplicated twice. line 15-24 & 27-36
try implementing a separate function to avoid this

ostapT (changes requested):

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+        while True:
+            content = input("Enter content line: ")
+            if content == "stop":
+                f.write("\n")
+                break
+            f.write(f"{line} {content}\n")
+            line += 1
+
+
+print(create_file())
```
YuriiKindrat, line 34: remove unnecessary print here, u can use `if __name__ == __main__`

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+def create_file() -> None:
+    if "-d" in sys.argv:
+        dir_path = os.path.join(*sys.argv[sys.argv.index("-d") + 1:])
+        os.makedirs(dir_path, exist_ok=True)
+    if "-f" in sys.argv:
+        file_name = sys.argv[sys.argv.index("-f") + 1]
+        if os.path.isfile(file_name):
+            writing_in_file()
+        else:
+            writing_in_file()
```
YuriiKindrat, line 16: why do u need this condition if u are making the same thing ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+import os
+import sys
+
+from datetime import datetime
+
+
+def create_file() -> None:
+    if "-d" in sys.argv:
+        dir_path = os.path.join(*sys.argv[sys.argv.index("-d") + 1:])
+        os.makedirs(dir_path, exist_ok=True)
```
wQuelS, line 10: what if dirs already exist?

wQuelS (changes requested):

PR #277, opened 2023-02-25
'Solution'

alexlega:


YuriiKindrat (changes requested):
Separate your code into functions to make it more modular

YuriiKindrat (approved):

PR #281, opened 2023-02-25
Solution

TeliksPG:


MikhailLyvak (approved):

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+            make_dir_and_file(command[3], command[4], command[6])
+        elif "-d" in command:
+            create_dir(command[3], command[4])
+        elif "-f" in command:
+            create_file(command[3])
+        elif terminal == "stop":
+            print("Exiting")
+            break
+        else:
+            print("Unsupported command!")
```
YuriiKindrat, line 44: what if command will have more/less than 2 dirs ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+            line = input("Enter content line: ")
+            if line == "stop":
+                file.write("\n")
+                break
+            file.write(line)
+            file.write("\n")
+
+
+def create_dir(direct: list) -> None:
+    direction = os.path.join("/".join(direct))
```
wQuelS, line 19: why do u use `/` here ? Read documentation about `os.path.join`. You should use it to ensure that path will be correct no matter what OS is running your program. It will automatically insert `/` or `\`. Fix everywhere

wQuelS (changes requested):

YuriiKindrat (approved):

PR #283, opened 2023-02-26
Solution new

destr34:


app/create_file.py
```diff
@@ -1 +1,81 @@
...
+        with open(file_path, "a") as file:
+            timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            file.write(f"\n{timestamp}\n")
+            count_old_file = 1
+            while True:
+                line = input("Enter content line: ")
+                if line == "stop":
+                    break
+                file.write(f"{count_old_file} {line}\n")
+                count_old_file += 1
```
olena-hul, line 28: Almost identical code in lines  31-40

app/create_file.py
```diff
@@ -1 +1,81 @@
...
+        os.makedirs(dir_path, exist_ok=True)
+        print("Directory created")
+    except Exception as e:
+        print(f"Error creating directory: {str(e)}")
+        sys.exit(1)
+
+
+def create_file(file_path: str) -> None:
+    if os.path.exists(file_path):
+        print("File already exist, you can add somthing")
```
olena-hul, line 17:
```suggestion
        print("File already exists, you can add something")
```

app/create_file.py
```diff
@@ -1 +1,81 @@
...
+
+def main() -> None:
+    args = sys.argv[1:]
+    dir_path = ""
+    file_path = ""
+
+    if "-d" in args and "-f" in args:
+        dir_index = args.index("-d")
+        file_index = args.index("-f")
+        path_parts = args[dir_index + 1:file_index]
```
olena-hul, line 54: -f flag can be passed first

olena-hul (changes requested):

app/create_file.py
ArsenPidhoretskyi: Not fixed you code will not work if -f flag will be first

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,80 @@
...
+        file_name = args[file_index]
+        file_path = os.path.join(dir_path, file_name)
+        create_file(file_path)
+
+    elif not args:
+        print(
+            "Invalid arguments. Usage: python create_file.py "
+            "[-d directory_path] [-f file_name]"
+        )
+        sys.exit(1)
```
Bohdan-Salamakha, line 76: separate your function into a few functions with own logic

Bohdan-Salamakha (changes requested):
v1

app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_path(dir_flag: (int, None),
+                file_flag: (int, None),
```
Bohdan-Salamakha, line 7: use typehint like: Optional[YOUR_VALUE_TYPE] = None
for example: list that contain string elements and can be None has typehint Optional[list[str]]

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+                args: list) -> str:
```
Bohdan-Salamakha, line 8: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+args = sys.argv
+dir_flag = args.index("-d") if "-d" in args else None
+file_flag = args.index("-f") if "-f" in args else None
+path = ""
+if dir_flag:
+    path = create_path(dir_flag, file_flag, args)
+if file_flag:
+    file_path = os.path.join(path, args[file_flag + 1])
+    file_content(file_path)
```
Bohdan-Salamakha, line 40: why did u remove main() function?
put this code block into main() and call it from program startpoint: if __name__ == "__main__": ...

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

PR #285, opened 2023-02-26
solution

AlessaQ:


app/create_file.py
```diff
@@ -1 +1,35 @@
...
+
+def terminal() -> None:
+    if "-d" in sys.argv and "-f" in sys.argv:
+        create_dir(sys.argv[sys.argv.index("-d") + 1:sys.argv.index("-f")])
+        create_file(sys.argv[-1])
+        return
+    if "-d" in sys.argv:
+        create_dir(sys.argv[2:])
+    if "-f" in sys.argv:
+        create_file(sys.argv[-1])
```
YuriiKindrat, line 14: no need to use blank return here, use "elif" instead

app/create_file.py
```diff
@@ -1 +1,35 @@
...
+
+
+def create_dir(path: list) -> None:
+    os.makedirs(os.getcwd() + "\\" + "\\".join(path))
```
YuriiKindrat, line 18: image
fix this

YuriiKindrat (changes requested):

app/create_file.py
AlessaQ: fixed

app/create_file.py
AlessaQ: Fixed. Also the create_directiry function has been moved to the terminal. Also, the creation of a file by calling `-f` and `-d` at the same time has been fixed (previously the file was not created in the directory)

YuriiKindrat (approved):

PR #287, opened 2023-02-26
Solution

oksana0410:


Anatolii-Poznyak (approved):

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+
+
+task = sys.argv[1:]
+
+
+def create_file_and_content(name: str) -> None:
+    with open(name, "a") as f:
+        content = []
+        content.append(datetime.now().strftime("%Y-%m-%d %X"))
+        i = 1
```
YuriiKindrat, line 13: "i" is a bad name for variable

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+
+        while True:
+            line = input("Enter content line: ")
+            if line == "stop":
+                break
```
YuriiKindrat, line 18: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+            content.append(str(i) + " " + line)
+            i += 1
+
+        f.writelines(content)
+
+
+if "-f" in task:
+    if "-d" in task:
+        path = os.path.join(*task[1:-2])
+        os.makedirs(path)
```
YuriiKindrat, line 28: what if dirs, already exists ?

YuriiKindrat (changes requested):

PR #294, opened 2023-02-27
Solution

VladyslavCherkashyn:


app/create_file.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import os
+from datetime import datetime
+import argparse
```
alexlega, line 3: tip: import modules firstly, then separate functions

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+parser = argparse.ArgumentParser(description="parsing command from terminal")
+parser.add_argument("-d", "--directories", type=str, nargs="+",
+                    help="creating directories")
+parser.add_argument("-f", "--file", type=str,
+                    help="creating directories")
+args = parser.parse_args()
+print(args.__dict__)
```
alexlega, line 11: redundant print

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+
+if args.__dict__["file"] is not None:
+    directory_path = os.getcwd()
+    if os.path.exists(temp):
+        directory_path = temp
+    file_name = args.__dict__["file"]
+    mode = "a" if os.path.exists(
+        os.path.join(directory_path, file_name)
+    ) else "w"
```
alexlega, line 28: If I understood all correctly, u don't need two regimes, because "a" and "w" create files if it doesn't exist

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+    with open(os.path.join(directory_path, file_name), mode) as file:
+        if mode == "a":
+            file.write("\n")
```
alexlega, line 32: According to my previous comment, I think u don't need it

alexlega (changes requested):

app/create_file.py
YuriiKindrat: @alexlega is right, fix this

app/create_file.py
YuriiKindrat: fix this

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+                    help="creating directories")
+args = parser.parse_args()
+print(args.__dict__)
+temp = ""
+if args.__dict__["directories"] is not None:
+
+    for directory in args.__dict__["directories"]:
+        temp = os.path.join(temp, directory)
+        if not os.path.exists(temp):
+            os.mkdir(temp)
```
YuriiKindrat, line 18: use make dirs instead to simplify ur code

app/create_file.py
YuriiKindrat: yeah, no need to make this condition here

YuriiKindrat (changes requested):
remove unnecessary file from ur PR

app/create_file.py
VladyslavCherkashyn: Sorry, but I don`t get how makedirs should simplify it. 
When i tried to refactor via makedirs - folder creation logic breaks if this path already exists

YuriiKindrat: @VladyslavCherkashyn u can use exist_ok param for makedirs func )

YuriiKindrat (changes requested):
file.txt is still in ur PR - remove it

app/create_file.py
VladyslavCherkashyn: exist_ok=True!!!!!!! for sure! Sorry, looks like I had a brain lag.

PR #297, opened 2023-02-27
Solution

myroslavatrschk:


app/create_file.py
```diff
@@ -1 +1,27 @@
...
+    if "-f" in command:
+        with open(command[-1], "w") as new_file:
+            new_file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+            line_number = 1
+            while True:
+                line_input = input("Enter content line: ")
+                if line_input == "stop":
+                    break
+                new_file.write(f"{line_number} {line_input} \n")
+                line_number += 1
```
YuriiKindrat, line 27: what if we will swap the flags, f.e. `python create_file.py  -f file.txt -d dir1 dir2`?
also, add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

YuriiKindrat (changes requested):
also, u didn't call this method anywhere, so it won't work after runnig

PR #298, opened 2023-02-28
Solution

Phoenix-Erazer:


app/create_file.py
```diff
@@ -1 +1,52 @@
...
+from sys import argv
+from os import path, makedirs
+from datetime import datetime
+
+
+def is_param(index: int, commands: list) -> bool:
+    return index < len(commands) and commands[index] not in ["-f", "-d"]
+
+
+def get_commands(commands: list) -> dict[str]:
```
ArsenPidhoretskyi, line 10: add type for commands. list of what
also for return type must be dict[str, str]

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #302, opened 2023-03-02
Solution

Rename925:


app/create_file.py
```diff
@@ -1 +1,34 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+count = 0
+home_directory = sys.argv
+if "-d" in home_directory:
+    if "-f" in home_directory:
+        file_path = "\\".join(
```
katevasilenko, line 9: Please use only os.path.join to create path all over the project

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+                           home_directory.index("-f")]
+        )
+        path = os.path.join(file_path)
+    else:
+        file_path = "\\".join(home_directory[home_directory.index("-d") + 1::])
+        path = os.path.join(file_path)
+    os.makedirs(path)
+
+if "-f" in home_directory:
+    if "-d" in home_directory:
```
katevasilenko, line 20: You are repeating yourself from the 7-8 lines

katevasilenko (changes requested):
You need to split your code to function or even several functions. You can for example create directory in first and create file in second. Check arguments once like if -f and -d, if only -f, if only -d.

vasylhnatiuk (changes requested):
Make sure that the code is not repeated (lines 19–25, 39–45), then it is better to make a function that will call other functions depending on the input data.

app/create_file.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import datetime
+import os
+import sys
+
+
+def file_from_terminal() -> None:
+    home_directory = sys.argv
```
Grayder0152, line 7: why name is `home_directory`?

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+            text = input("Enter content line: ")
+            if text == "stop":
+                break
+            file.write(f"{count} {text}\n")
+
+
+def create_directory(directories: list) -> None:
+    for directory in directories:
+        os.mkdir(directory)
+        os.chdir(directory)
```
Grayder0152, line 37: use `os.makedirs` method

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+import os
+import sys
+
+
+def file_from_terminal() -> None:
+    home_directory = sys.argv
+
+    if "-d" in home_directory and "-f" in home_directory:
+        create_directory(home_directory[2:-2])
+        create_file(home_directory[-1])
```
Grayder0152, line 11: what if `-f` before `-d`?

Grayder0152 (changes requested):

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+            text = input("Enter content line: ")
+            if text == "stop":
+                break
+            file.write(f"{count} {text}\n")
+
+
+def create_directory(directories: list) -> None:
+    for directory in directories:
+        os.makedirs(directory)
+        os.chdir(directory)
```
ArsenPidhoretskyi, line 37: No need this
dust do:
os.makedirs(os.path.join(*directories), exist_ok=True)

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+import datetime
+import os
+import sys
+
+
+def file_from_terminal() -> None:
+    directory = sys.argv
+
+    if "-d" in directory and "-f" in directory:
+        create_directory(directory[2:-2])
```
ArsenPidhoretskyi, line 10: Why from 2 to -2
Check if your code works on two cases
python create_file.py -f file.txt -d dir1
python create_file.py -d dir1 -f file.txt

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    if "-d" in directory and "-f" in directory:
+        check_variable = " ".join(directory)
+        if check_variable.index("-d") < check_variable.index("-f"):
+            os.chdir(create_directory(directory[2:-2]))
+            create_file(directory[-1])
+            return
+        else:
+            create_directory(directory[directory.index("-d") + 1::])
+            create_file(directory[directory.index("-f") + 1])
+            return
```
ArsenPidhoretskyi, line 17: YOu can use this instead of 12-13 line move it after if-else block

ArsenPidhoretskyi:
```
if check_variable.index("-d") < check_variable.index("-f"):
    os.chdir(create_directory(directory[2:-2]))
else:
    create_directory(directory[directory.index("-d") + 1::])

create_file(directory[directory.index("-f") + 1])
return
```

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #303, opened 2023-03-03
Solution

ochernyi:


app/create_file.py
```diff
@@ -1 +1,26 @@
...
+import sys
+
+command = sys.argv
+
+
+def create_file() -> None:
+    if "-d" in command:
+        index = 2
+        while len(command) != 2 and command[index] != "-f":
+            os.mkdir(command[index])
```
vasylhnatiuk, line 13: Use the function os.makedirs() with the parameter exist_ok=True to avoid an error if the directory already exists

app/create_file.py
```diff
@@ -1 +1,26 @@
...
+            os.chdir(command[index])
+            index += 1
+
+    if "-f" in command:
+        with open(command[-1], "w") as new_file:
+            new_file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+            line_number = 1
+            while True:
```
vasylhnatiuk, line 21: Replace `while True:` with `while line_input != "stop":` for better code clarity:

vasylhnatiuk (changes requested):

app/create_file.py
```diff
@@ -1 +1,24 @@
...
+            index += 1
+
+    if "-f" in command:
+        with open(command[-1], "w") as new_file:
+            new_file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+            line_number = 1
+            line_input = input("Enter content line: ")
+            while line_input != "stop":
+                new_file.write(f"{line_number} {line_input} \n")
+                line_number += 1
```
Grayder0152, line 24: it's an infinite loop

Grayder0152 (approved):

app/create_file.py
ochernyi: > it's an infinite loop

Yes, because the task demands it. 
(After pressing Enter it creates file file.txt and then terminal should ask you to input content lines until you input "stop":)

vasylhnatiuk (approved):
Good job!

PR #304, opened 2023-03-05
Solution

StreetNik:


vasylhnatiuk (changes requested):
it is a not bad solution, but try to change solution and divide code into a few functions each of which will be responsible for a single functional. use minimum `while` loops. and after this run functions, example:
```
 if __name__ == "__main__":
    main()
```

app/create_file.py
```diff
@@ -3,30 +3,45 @@
 from datetime import datetime
 
 
-terminal_command = sys.argv[1:]
 index = 0
```
StreetNik, line 6: I am not sure what is better, whether I have to do it using global or just pass(index) to function as a parameter

Vasyl-Smutok (changes requested):
Please fix `flake8` error

StreetNik:
> Please fix `flake8` error
the problem here is I use:
if __name__ == "__main__":
    main()
To be honest I don't know how to automatically call functions without this structure

vasylhnatiuk (approved):
Good job!

PR #307, opened 2023-03-08
Solution

arsenmakovei:


app/create_file.py
```diff
@@ -1 +1,41 @@
...
+def create_file() -> None:
+    commands = sys.argv
+
+    if "-d" in commands and "-f" in commands:
+        dirs_path = "/".join(
+            command for command in commands[2:-1] if command != "-f"
+        )
+        os.makedirs(dirs_path, exist_ok=True)
+
+        path_to_file = dirs_path + "/" + commands[-1]
```
Grayder0152, line 28: the same as above

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+            f.write(str(line_number) + " " + content + "\n")
+            content = input("Enter content line: ")
+            line_number += 1
+
+
+def create_file() -> None:
+    commands = sys.argv
+
+    if "-d" in commands and "-f" in commands:
+        dirs_path = "/".join(
```
Grayder0152, line 23: use `os.path.join`

Grayder0152 (changes requested):

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+            f.write(str(line_number) + " " + content + "\n")
+            content = input("Enter content line: ")
+            line_number += 1
+
+
+def create_file() -> None:
+    commands = sys.argv
+
+    if "-d" in commands and "-f" in commands:
+        dirs_path = os.path.join(*commands[2:-2])
```
katevasilenko, line 23: It would be better and more readable to use index() to find place of -f and -d flags and work it’s it instead of just give [2:-2] slice

katevasilenko (changes requested):

vasylhnatiuk (approved):
Well done!

PR #308, opened 2023-03-09
Solution

lotrik84:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+import os
+import argparse
+
+from datetime import datetime
+
+
+def create_directories(directories: list) -> str:
+    path_to_file = os.path.join(*directories)
+    if not os.path.exists(path_to_file):
+        os.makedirs(path_to_file)
```
Grayder0152, line 10:
```suggestion
    os.makedirs(path_to_file, exist_ok=True)
```

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+
+    return path_to_file
+
+
+def create_file(path: str, filename: str, lines: str) -> None:
+    path = os.path.join(path + "/" + filename)
```
Grayder0152, line 16: `/` will add after join

Grayder0152 (changes requested):

app/create_file.py
lotrik84: no, it doesn't) I've checked.

Grayder0152: image

Grayder0152 (changes requested):

vasylhnatiuk (approved):
Well done!

PR #309, opened 2023-03-09
Add solution

ZaikaDaria:


app/create_file.py
```diff
@@ -1 +1,26 @@
...
+import datetime
+import os
+import sys
+
+
+def create_file() -> None:
+    if "-d" in sys.argv and "-f" in sys.argv:
+        dir_name = "/".join(
+            sys.argv[sys.argv.index("-d") + 1: sys.argv.index("-f")]
+        )
```
ArsenPidhoretskyi, line 10: Use os.path.join

app/create_file.py
```diff
@@ -1 +1,26 @@
...
+        file_name = sys.argv[sys.argv.index("-f") + 1]
+        path = f"{os.getcwd()}/{dir_name}/{file_name}"
```
ArsenPidhoretskyi, line 12: Use os.path.join

ArsenPidhoretskyi (changes requested):
> If only -d flag passed, means all items after this flag are parts of the path.
> python create_file.py -d dir1 dir2 - creates directory dir1/dir2 inside current directory. 
> If only -f flag passed, means first item is the file name.
Does not covered

Vasyl-Smutok (approved):
Good Job 👍

PR #310, opened 2023-03-10
Solution

Roman28101:


app/create_file.py
```diff
@@ -1 +1,42 @@
...
+    if "-d" not in terminal_command:
+        file_name = terminal_command[terminal_command.index("-f") + 1]
+        create_file(file_name)
+    if "-f" not in terminal_command:
+        dirs = terminal_command[(terminal_command.index("-d") + 1):]
+        create_dirs(dirs)
+    if "-d" in terminal_command and "-f" in terminal_command:
+        file_name = terminal_command[terminal_command.index("-f") + 1]
+        dirs = terminal_command[
+            (terminal_command.index("-d") + 1):terminal_command.index("-f")]
```
ArsenPidhoretskyi, line 37: Not true because -f parameter can be first

app/create_file.py
```diff
@@ -1 +1,42 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_dirs(dirs: list) -> str:
+    path = os.path.join(*dirs)
+    if not os.path.exists(path):
+        os.makedirs(path)
```
ArsenPidhoretskyi, line 9: You can use os.makedirs(path, exist_ok=True)

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #311, opened 2023-03-11
Solution

Sashkoooo:


app/create_file.py
```diff
@@ -1 +1,56 @@
...
+
+
+def add_timestamp(path: str) -> None:
+    now = datetime.now()
+    timestamp = now.strftime("%Y-%m-%d %H:%M:%S")
+    with open(path, "a") as file:
+        file.write(f"{timestamp}\n")
+
+
+terminal_parser = argparse.ArgumentParser()
```
katevasilenko, line 33: Please use `if __name__ == "__main__": `

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+terminal_parser.add_argument(
+    "-d", dest="directories", type=str, nargs="*", metavar="directories names",
+    help="creates directories inside current directory"
+)
+terminal_parser.add_argument(
+    "-f", dest="file_name", type=str, nargs=1, metavar="file name",
+    help="create file and input content lines until 'stop'"
+)
```
katevasilenko, line 41: Also you can create function for parsing arguments as well but it’s up  to you

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+    current_working_directory = create_path(argument_in_terminal.directories)
+    create_directory(current_working_directory)
+
+if argument_in_terminal.file_name:
+    path_to_file = [
+        current_working_directory, argument_in_terminal.file_name[0]
+    ]
+    path = create_path(path_to_file)
+    add_timestamp(path)
+    content_input(path)
```
katevasilenko, line 56: For me it is looks useless to open file twice firstly to write timestamp and second time to write all other content. Does that make sense to combine those function to one? Or also you can return prepared timestamp from first function and call it in second one to write this value to a opened file

katevasilenko (changes requested):

Vasyl-Smutok (approved):
Great 👍

PR #313, opened 2023-03-13
Develop

andy77andy:


app/create_file.py
```diff
@@ -1 +1,37 @@
...
+from datetime import datetime
+import os
+import sys
+
+
+def create_file_from_terminal() -> None:
+    command = sys.argv
+    path = None
+    if "-f" == command[1]:
+        path = os.path.join(str(os.getcwd()), command[-1])
```
ArsenPidhoretskyi, line 10: It is not true because if -d in command than command[-1] will equal to dirictory

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+    else:
+        dir_path = (
+            os.path.join(*command[2:-2])
+            if "-f" == command[-2]
+            else os.path.join(*command[2:]))
```
ArsenPidhoretskyi, line 15:
```suggestion
            os.path.join(*command[2:-2])
            if "-f" == command[-2] else 
            os.path.join(*command[2:])
        )
```

ArsenPidhoretskyi (changes requested):

Vasyl-Smutok (approved):
good job 👍

PR #319, opened 2023-03-22
solution

DmytroProkofiev:


app/create_file.py
```diff
@@ -1 +1,36 @@
...
+                break
+            result_file.write(f"{line_number} {content_line}\n")
+            line_number += 1
+
+
+def full_creation() -> None:
+    input_lst = sys.argv
+    if "-d" in input_lst and "-f" in input_lst:
+        file_name = input_lst[input_lst.index("-f") + 1]
+        paths = input_lst[(input_lst.index("-d") + 1):input_lst.index("-f")]
```
olena-hul, line 29: -f flag can be passed first

`-f file1 -d dir1`

olena-hul (changes requested):

app/create_file.py
olena-hul: My comment was related to line  `paths = input_lst[(input_lst.index("-d") + 1):input_lst.index("-f")]`

For example if we have `-f file1 -d dir1` arguments in command line
Your code will do the following

`input_lst.index("-d") + 1` - this will return 4
`input_lst.index("-f")` - this will return 1

`input_lst[4:1]` will return `[]`

olena-hul (changes requested):

app/create_file.py
```diff
@@ -1 +1,36 @@
...
+                break
+            result_file.write(f"{line_number} {content_line}\n")
+            line_number += 1
+
+
+def full_creation() -> None:
+    input_lst = sys.argv
+    if "-d" in input_lst and "-f" in input_lst:
+        file_name = input_lst[input_lst.index("-f") + 1]
+        paths = input_lst[(input_lst.index("-f") + 1):input_lst.index("-d")]
```
olena-hul, line 29: Bad trick) You should cover both cases, such as
`-f file -d dir`
and
`-d dir -f file`

olena-hul: Just replacing -f and -d flags in line 29 doesn't fix the problem))))

olena-hul (changes requested):

olena-hul (approved):

olena-hul:
Good job!

PR #325, opened 2023-04-04
Solution

anastasia-martyniuk:


app/create_file.py
```diff
@@ -1 +1,42 @@
...
+
+            while True:
+                text = input("Enter content line: ")
+
+                if text == "stop":
+                    break
+                file.write(f"{line_number} {text}\n")
+                line_number += 1
+
+            file.write("\n")
```
EdAlekseiev, line 38: Separate this function into several functions. For example, create a function for creating directories and a function for creating files.

EdAlekseiev (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+        file.write("\n")
+
+
+def main() -> None:
+    file_name = None
+
+    text_from_terminal = sys.argv
+
+    if "-f" in text_from_terminal:
+        file_name = text_from_terminal[-1]
```
ArsenPidhoretskyi, line 35: Why?? Is always filename will be last

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+        text_from_terminal = text_from_terminal[:-2]
+
+    if "-d" in text_from_terminal:
+        dirs_name = os.path.join(*text_from_terminal[2:])
```
ArsenPidhoretskyi, line 39: Why from second to end?
dirs_name = "-f/file.txt/dir1/dir"

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job! But also I suggest to not do os.chdir. It is not so good

PR #329, opened 2023-04-08
'Py-create-file-from-terminal Solution'

MilManIT:


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+            break
+        lines.append(line)
+
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    lines = [f"{i} {line}" for i, line in enumerate(lines, start=1)]
+    content = f"{timestamp}\n" + "\n".join(lines) + "\n"
+
+    mode = "a" if os.path.isfile(file_path) else "w"
+    with open(file_path, mode) as f:
+        f.write(content)
```
ArsenPidhoretskyi, line 33: Make it as function

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+import sys
+import os
+from datetime import datetime
+
+args = sys.argv[1:]
+dir_path = ""
+file_name = ""
+file_content = ""
+
+if "dir" in args:
```
ArsenPidhoretskyi, line 10: Why you check by "dir"?
We have two flags: -f and -d

ArsenPidhoretskyi (changes requested):
Task is not properly done

app/create_file.py
MilManIT: Because Olena Hul told me to do this

PR #334, opened 2023-04-10
'Solution'

HalynaPetrova:


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+    if "-d" in command:
+        if len(command) == 4:
+            add_dir = os.path.join(command[-2], command[-1])
+            os.makedirs(add_dir, exist_ok=True)
+
+        if len(command) == 6:
+            add_dir = os.path.join(command[-4], command[-3])
+            os.makedirs(add_dir, exist_ok=True)
+            way = os.path.join(command[-4], command[-3], command[-1])
+            add_content(way)
```
Vasyl-Smutok, line 33: what will happen in this case?
```
python create_file.py -d dir1 dir2 dir3 -f file.txt
```

Vasyl-Smutok (changes requested):

app/create_file.py
```diff
@@ -1 +1,34 @@
...
+                break
+
+
+def create_file() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-f", type=str)
+    parser.add_argument("-d", type=str, nargs="*")
+    args = parser.parse_args()
+
+    if args.d is None:
```
ArsenPidhoretskyi, line 24: But if -f is None also?
Maybe you first need to check if -d is not None
and than if -f is not None

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #343, opened 2023-04-20
Solution

sasha-semenenko:


app/create_file.py
```diff
@@ -1 +1,39 @@
...
+from datetime import datetime
+
+
+def create_file() -> None:
+    command = sys.argv
+
+    if "-d" in command:
+        current_directory = command[command.index("-d") + 1]
+        path = os.path.join(
+            current_directory, command[command.index("-d") + 2]
```
olena-hul, line 12: Why do you use +2?

olena-hul (changes requested):

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+def create_file() -> None:
+    commands = sys.argv
+
+    file_path = get_directory(commands)
+
+    if file_path is not None:
+        create_and_write_file(file_path)
+
+
+create_file()
```
YukihiroSM, line 56:
```suggestion
if __name__ == "__main__":
    create_file()
```
It's better to perform this in the way like aboe, because if you will try to import this module - you will start the function itself.

YukihiroSM (approved):

PR #347, opened 2023-04-25
solution

vborovets0:


Bohdan-Salamakha (changes requested):
I should be able to use command any number of times
`python create_file.py -d dir1 dir2 -f file.txt`
image

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+
+        f.write(str(datetime.datetime.now().
+                    strftime("%Y-%m-%d %H:%M:%S")) + "\n")
+        text = input("Enter content line: ")
+        count = 0
+        while text != "stop":
+            count += 1
+            f.write(f"{count} {text} \n")
+            text = input("Enter content line: ")
+        f.write("\n")
```
Bohdan-Salamakha, line 31: use more declarative name than "f"

Bohdan-Salamakha (changes requested):
v1

PR #349, opened 2023-05-01
Solution

svyshcho:


app/create_file.py
```diff
@@ -1 +1,38 @@
-# write your code here
+import os
+import sys
+
+from datetime import datetime
+
+
+def create_file() -> None:
```
EdAlekseiev, line 7: Please, separate this function into a few small functions.

svyshcho: Done

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+        os.makedirs(path, exist_ok=True)
+        file_path = os.path.join(path, command[-1])
+        open_file(file_path)
+
+    if command[1] == "-d" and "-f" not in command:
+        os.makedirs(os.path.join(*command[2:]))
+
+
+def open_file(command: str) -> None:
+    with open(command, "w") as file:
```
EdAlekseiev, line 26: If the file exists app should add content below. App rewrites file content now.

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+        open_file(command[-1])
+
+    if command[1] == "-d" and "-f" in command:
+        path = os.path.join(*command[2:-2])
+        os.makedirs(path, exist_ok=True)
+        file_path = os.path.join(path, command[-1])
+        open_file(file_path)
+
+    if command[1] == "-d" and "-f" not in command:
+        os.makedirs(os.path.join(*command[2:]))
```
EdAlekseiev, line 22: It does not work for `python create_file.py -f file.txt -d dir1 dir2`

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+import os
+import sys
+
+from datetime import datetime
+
+
+command = sys.argv
+
+
+def create_file(command: list) -> None:
```
EdAlekseiev, line 10: What happens if the command will be empty?

EdAlekseiev (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #351, opened 2023-05-02
Solution

Fawnygod:


app/create_file.py
```diff
@@ -1 +1,38 @@
...
+import sys
+from datetime import datetime
+
+
+argv = sys.argv
+now = datetime.now()
+
+
+def create_folder() -> None:
+    dirs = argv[argv.index("-d") + 1:argv.index("-f")] \
```
ArsenPidhoretskyi, line 11: But if -f flag will be first it will not work

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+        if "-f" in argv else argv[argv.index("-d") + 1:]
+    os.makedirs(os.path.join(*dirs), exist_ok=True)
+
+
+def create_file() -> None:
+    if "-d" in argv:
+        folders = argv[argv.index("-d") + 1:argv.index("-f")]
```
ArsenPidhoretskyi, line 18: reuse path from create_folder

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #356, opened 2023-05-12
Solution

OlehOryshchuk:


app/create_file.py
```diff
@@ -1 +1,54 @@
...
+
+
+def create_file(file_path: str) -> None:
+    if os.path.exists(file_path):
+        with open(file_path, "a") as file:
+            file.write("\n" + add_content())
+
+    else:
+        with open(file_path, "w") as new_file:
+            new_file.write(add_content())
```
EdAlekseiev, line 27: You can open the file in mode "a" in each case

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+    date_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    content = date_time + "\n"
+
+    while True:
+        user_text = input("Enter content line: ")
+        if user_text == "stop":
+            return content
+
+        content += f"{line} {user_text}\n"
+        line += 1
```
EdAlekseiev, line 17: It's not a good idea to store all entered strings in the variable.  Put on content to file at once.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #359, opened 2023-05-15
solution

T-Vlad-96:


app/create_file.py
```diff
@@ -1 +1,41 @@
...
+        content: list
+) -> None:
+    with open(file_name, mode) as f:
+        f.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        f.writelines(content)
+
+
+def create_file() -> None:
+    command_line = sys.argv
+    if command_line[1] == "-f":
```
olena-hul, line 28: `-f` flag can be everywhere, not exactly at index 1. Use `.index()` function to determine the index of the flag

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+        content = content_for_new_file()
+        write_content_in_new_file(command_line[-1], "w", content)
+    elif command_line[1] == "-d":
+        os.makedirs(os.path.join(*command_line[2::]))
+    else:
+        os.makedirs(os.path.join(*command_line[2:-1]))
+        os.chdir(os.path.join(*command_line[2:-1]))
+        content = content_for_new_file()
+        write_content_in_new_file(command_line[-1], "w", content)
```
olena-hul, line 37: It won't work correctly. If the first argument is `-d` and then goes `-f`, the `elif` block will be excuted, not `else`. You should change your conditions.

Keep in mind, that the order of the flags can be different. For example, `-f` can go first, and `-d` can go after it

olena-hul (changes requested):

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+    while True:
+        line = input()
+        if line == "stop":
+            break
+        content.append(line + "\n")
+    return content
+
+
+def write_content_into_new_file(file_name: str, content: list) -> None:
+    with open(file_name, "w") as f:
```
YukihiroSM, line 17: It rewrites the file content every time, but it should append new data to the file.

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+                    *command_line[command_line.index("-d") + 1::]
+                )
+            os.makedirs(path_)
+            os.chdir(path_)
+        file_name = command_line[command_line.index("-f") + 1]
+        content = create_content_for_new_file()
+        write_content_into_new_file(file_name, content)
+    elif "-d" in command_line:
+        path_ = os.path.join(*command_line[command_line.index("-d") + 1::])
+        os.makedirs(path_)
```
YukihiroSM, line 42: What if directory already exists?

YukihiroSM (changes requested):

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+    with open(file_name, "a") as f:
+        f.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        f.writelines(content)
+
+
+def create_directory_with_path(path_: str) -> None:
+    try:
+        os.makedirs(path_)
+    except FileExistsError:
+        pass
```
EdAlekseiev, line 28:
```suggestion
    os.makedirs(path_, exist_ok =True)
```

EdAlekseiev (changes requested):

app/create_file.py
T-Vlad-96: thanks

olena-hul:
Good job!

PR #368, opened 2023-05-23
solution

antonpiddubnyi:


app/create_file.py
```diff
@@ -1 +1,34 @@
...
+                file.write("\n")
+                break
+            file.write(f"{line_number} {line_text}\n")
+
+
+def creating_file_from_command() -> None:
+    if "-f" in command:
+        if "-d" in command:
+            command.pop(command.index("-f"))
+            os.makedirs(os.path.join(*command[2:-1]), exist_ok=True)
```
ArsenPidhoretskyi, line 25: But if -f flag will be first???
Also this case must works the same

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+            if line_text == "stop":
+                file.write("\n")
+                break
+            file.write(f"{line_number} {line_text}\n")
+
+
+def creating_file_from_command() -> None:
+    if "-f" in command:
+        if "-d" in command:
+            if command.index("-f") < command.index("-d"):
```
ArsenPidhoretskyi, line 24: As for me must be:
```suggestion
            if command.index("-f") > command.index("-d"):
```
Can you pin screenshots with your testing?

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+def creating_file_from_command() -> None:
+    if "-f" in command:
+        if "-d" in command:
+            if command.index("-f") > command.index("-d"):
+                command.pop(command.index("-f"))
+                os.makedirs(os.path.join(*command[2:-1]), exist_ok=True)
+                create_file(os.path.join(*command[2:]))
+            else:
+                command.pop(command.index("-f"))
+                os.makedirs(os.path.join(*command[3:]), exist_ok=True)
```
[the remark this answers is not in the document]
antonpiddubnyi: Arsen, I don't get it.  Do you want the path to be the same in both cases like you suggested in the above comments or do you want them to be different depending on which flag is passed first?

ArsenPidhoretskyi (approved):
Well done!

PR #370, opened 2023-05-24
Py create file from terminal

alexgubarev88:


app/create_file.py
```diff
@@ -1 +1,32 @@
...
+import sys
+from datetime import datetime
+
+
+console_data = sys.argv
+
+if "-d" in console_data:
+    parent_dir = console_data[console_data.index("-d") + 1]
+    path = os.path.join(parent_dir, console_data[console_data.index("-d") + 2])
+    os.makedirs(path)
```
Grayder0152, line 11: what if path already exist ? add `exist_ok=True` param

app/create_file.py
```diff
@@ -1 +1,32 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+console_data = sys.argv
+
+if "-d" in console_data:
+    parent_dir = console_data[console_data.index("-d") + 1]
```
Grayder0152, line 9: save to variable `console_data.index("-d") + 1` to avoid dublication of code in the next lines

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+
+    with open(file_path, "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+        counter = 0
+
+        while True:
+            message = input("Enter content line: ")
+            counter += 1
+            if message == "stop":
+                break
```
Grayder0152, line 31: add endline before closing file

app/create_file.py
```diff
@@ -1 +1,32 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+console_data = sys.argv
+
```
Grayder0152, line 7: better to create function for those code

Grayder0152 (changes requested):

app/create_file.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory(console_data):
+    parent_dir = console_data[console_data.index("-d") + 1]
+    path = os.path.join(parent_dir, console_data[console_data.index("-d") + 2])
```
vsmutok, line 8: why `+2`, if, for example, we want to create 5 or 10 folders
```suggestion
    path = os.path.join(parent_dir, console_data[console_data.index("-d") + ???])
```

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+        while True:
+            message = input("Enter content line: ")
+            counter += 1
+            if message == "stop":
+                break
+            file.write(f"{counter} {message}\n")
+        file.write("\n")
+
+
```
vsmutok, line 33: add

```python
if __name__ == '__main__':
```

vsmutok (changes requested):

PR #372, opened 2023-05-24
Py create file from terminal

alexgubarev88:


app/create_file.py
```diff
@@ -1 +1,46 @@
...
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory(console_data: list[str]) -> None:
+    parent_dir: str = console_data[console_data.index("-d") + 1]
+
+    dir_indices = [i for i, arg in enumerate(console_data) if arg == "-d"]
+    dir_names = [console_data[i + 1] for i in dir_indices]
```
ArsenPidhoretskyi, line 10: Why so hard code??????
Why you can not use slices?

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+    for dir_name in dir_names:
```
ArsenPidhoretskyi, line 12: Loop is redundant in this case
You can do: os.path.join(*list_of_dirs)

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+        path: str = os.path.join(parent_dir, dir_name)
+        os.makedirs(path, exist_ok=True)
+
+
+def create_file(console_data: list[str]) -> None:
+    if "-d" in console_data:
+        parent_dir: str = console_data[console_data.index("-d") + 1]
+        new_dir: str = console_data[console_data.index("-d") + 2]
```
ArsenPidhoretskyi, line 20: So strange code. Also will not work in this case:

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory(console_data: list[str]) -> None:
+    parent_dir: str = console_data[console_data.index("-d") + 1]
+    dir_names = console_data[console_data.index("-d") + 2:]
```
ArsenPidhoretskyi, line 8: the same as:
dir_names = console_data[console_data.index("-d") + 1:]

ArsenPidhoretskyi: But if -f flag be after -d flag?

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+
+
+def create_file(console_data: list[str]) -> None:
+    if "-d" in console_data:
+        parent_dir: str = console_data[console_data.index("-d") + 1]
+        dir_indices = [i for i, arg in enumerate(console_data) if arg == "-d"]
+        dir_names: list[str] = []
+        for index in dir_indices:
+            if index + 1 < len(console_data):
+                dir_names.append(console_data[index + 1])
```
ArsenPidhoretskyi, line 21: Use slices

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory(console_data: list[str]) -> None:
+    dir_index = console_data.index("-d") + 1
+    parent_dir = console_data[dir_index]
+    dir_names = console_data[dir_index + 1:]
```
ArsenPidhoretskyi, line 9: Your code not works with the next command:

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+    path = os.path.join(parent_dir, *dir_names)
+    os.makedirs(path, exist_ok=True)
+
+
+def create_file(console_data: list[str]) -> None:
+    if "-d" in console_data:
+        dir_index = console_data.index("-d") + 1
+        parent_dir = console_data[dir_index]
+        dir_names = console_data[dir_index + 1: console_data.index("-f")]
```
ArsenPidhoretskyi, line 19: But if -f flag will be first???

vasylhnatiuk (approved):

PR #375, opened 2023-05-26
Solution

vlamuha:


app/create_file.py
```diff
@@ -1 +1,37 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+arguments = sys.argv[1:]
+name = "tittle_was_not_supplied"
+directory = ["directory", "was", "not", "supplied"]
```
ArsenPidhoretskyi, line 8: So bad..... No need this default values
If you want default it must be empty list or None or ""

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+    f_index = arguments.index("-f")
+    name = arguments[f_index + 1]
+elif "-d" in arguments:
+    d_index = arguments.index("-d")
+    directory = arguments[d_index + 1:]
+
+path = os.path.join(*directory)
+os.makedirs(path, exist_ok=True)
+
+with open(path + "\\" + name, "a+") as f:
```
ArsenPidhoretskyi, line 28: use os.path.join

ArsenPidhoretskyi: rename f to file. Using one letter as name is not okay

ArsenPidhoretskyi (changes requested):
Better to write code as function

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+    while True:
+        message = input("Enter content line: ") + "\n"
+        if message == "stop\n":
+            break
+        text.write(message)
+    text.write("\n")
+
+
+def create_dic(file_name: list) -> str:
```
ArsenPidhoretskyi, line 19: create_dic is not so good names:
rename to create_directories

ArsenPidhoretskyi: file_name -> dir_names: list[str]

app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+def file_write(*filename) -> None:
```
ArsenPidhoretskyi, line 6: provide annotations

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+def file_write(*filename: str) -> None:
```
vsmutok, line 6: `str` wrong annotation

vsmutok (changes requested):
Also please use `black`  to fix indentation

ArsenPidhoretskyi (approved):
Well done!

PR #376, opened 2023-05-26
Solution

OksanaAdamchuk:


vsmutok (changes requested):
when I use `python create_file.py -d dir1 dir2` it creates a file for me, why?
should create only folders

OksanaAdamchuk:
> when I use `python create_file.py -d dir1 dir2` it creates a file for me, why? should create only folders

Because I thought that it means that if flag -f isn't in terminal, file should be created with default name file.txt, and not with the one mentioned after -f flag?

vsmutok:
if there is no `-f` the file should not be created at all

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+import os
+from datetime import datetime
+
+
+def create_file() -> None:
+    arguments = sys.argv[1:]
+    file_name = "file.txt"
+    directories = []
+    if "-f" in arguments:
+        if "-d" in arguments and "-f" in arguments:
```
vsmutok, line 11: you already checked it 2 lines above
```suggestion
        if "-d" in arguments:
```

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+            index_d = arguments.index("-d")
+            index_f = arguments.index("-f")
+            directories = arguments[(index_d + 1) : index_f]
+            file_name = arguments[-1]
+        if "-f" in arguments and "-d" not in arguments:
+            file_name = arguments[-1]
```
vsmutok, line 17:
```suggestion
        file_name = arguments[-1]
```

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+
+        directories_path = os.path.join(*directories) if directories else ""
+        file_name = str(file_name)
+        destination = os.path.join(directories_path, file_name)
+        if directories_path != "":
+            if not os.path.exists(directories_path):
+                os.makedirs(directories_path)
```
vsmutok, line 24:
```suggestion
            os.makedirs(directories_path, exist_ok=True)
```

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+            now = datetime.now()
+            date_time = now.strftime("%Y-%m-%d %H:%M:%S")
+            content.insert(0, date_time)
+            content.append("")
+            destination_file.write("\n".join(content))
+    else:
+        directories = arguments[(arguments.index("-d") + 1) :]
+        directories_path = os.path.join(*directories)
+        if not os.path.exists(directories_path):
+            os.makedirs(directories_path)
```
vsmutok, line 50:
```suggestion
        os.makedirs(directories_path, exist_ok=True)
```

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+            if not os.path.exists(directories_path):
+                os.makedirs(directories_path)
+
+        with open(destination, "a+") as destination_file:
+            content = []
+            line = ""
+            counter = 1
+            while line != "stop":
+                line = input("Enter content line: ")
+                if line != "stop":
```
vsmutok, line 32: you check it in while

vsmutok (changes requested):

Abnormaltype (approved):

PR #378, opened 2023-05-29
It works but I cry looking at its ugliness

N1khto:


vasylhnatiuk (changes requested):
Please try to separate your code into a few functions.

app/create_file.py
```diff
@@ -1 +1,43 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def define_path(args: list) -> str:
```
ArsenPidhoretskyi, line 6: It returns Optional[str]
Because if -d is not in args it will return None

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    if "-d" in args:
+        if "-f" in args:
+            return os.path.join(*args[args.index("-d") + 1:args.index("-f")])
```
ArsenPidhoretskyi, line 9: But if -f will be first?

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+        return os.path.join(*args[args.index("-d") + 1:])
+
+
+def define_file_name(args: list, path: str) -> str:
+    if path:
+        return os.path.join(path, args[-1])
+    return args[-1]
```
ArsenPidhoretskyi, line 16: Not true. You must always get from args it goes after -f flag

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #379, opened 2023-05-30
Solution

Munkoor:


app/create_file.py
```diff
@@ -1 +1,95 @@
...
+        create_file(file_path)
+        lines = []
+        while True:
+            line = input("Enter content line: ")
+            if line.lower() == "stop":
+                break
+            lines.append(line)
+        append_content(file_path, lines)
+    else:
+        print("Invalid arguments.")
```
ArsenPidhoretskyi, line 91: Task is hardcoded for indexes:
args[0] == "-d" and args[2] == "-f"
Can be false:
Use `in` and `.index` instead of hardcoded indexes

app/create_file.py
```diff
@@ -1 +1,95 @@
...
+
+def create_directory(path: str) -> None:
+    try:
+        os.makedirs(path)
+        print(f"Created directory: {path}")
+    except FileExistsError:
+        print(f"Directory already exists: {path}")
+    except OSError as e:
+        print(f"Failed to create directory: {path}")
+        print(f"Error: {str(e)}")
```
ArsenPidhoretskyi, line 14: Just simplify code with:
```suggestion
        os.makedirs(path, exist_ok=True)
        print(f"Created directory: {path}")
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
image
fix flake8 test

vasylhnatiuk (approved):
Good job

PR #381, opened 2023-06-02
1st version

romanfedorchuk:


app/create_file.py
```diff
@@ -1 +1,36 @@
...
+        f.write(f"{timestamp}\n")
+        while True:
+            line = input("Enter content line:")
+            if line == "stop":
+                break
+            f.write(f"{line}\n")
+
+
+if "-d" in path_list and "-f" not in path_list:
+    folders = path_list[2:]
```
olena-hul, line 28: Be consistent and use path_list[path_list.index("-d") + 1:] like you did for files

app/create_file.py
```diff
@@ -1 +1,36 @@
...
+import sys
+import os
+from datetime import datetime as dt
+
+
+path_list = sys.argv
+now = dt.now()
+timestamp = now.strftime("%Y-%m-%d %H:%M:%S")
+new_file = path_list[path_list.index("-f") + 1]
+folders = path_list[2: path_list.index("-f")]
```
olena-hul, line 10: What if I swap the order of indexes?
for example: `-f file1 -d dir1`

You should handle both cases

olena-hul (changes requested):

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+import sys
+import os
+from datetime import datetime as dt
+
+
+path_list = sys.argv
+now = dt.now()
+timestamp = now.strftime("%Y-%m-%d %H:%M:%S")
+new_file = path_list[path_list.index("-f") + 1]
+folders = path_list[path_list.index("-d") + 1:]
```
olena-hul, line 10: No need to define folders here, because you redefine them in `if` blocks

app/create_file.py
```diff
@@ -1 +1,39 @@
...
+            if line == "stop":
+                break
+            f.write(f"{line}\n")
+
+
+if "-d" in path_list and "-f" not in path_list:
+    folders = path_list[path_list.index("-d") + 1:]
+    create_folders()
+elif "-d" not in path_list and "-f" in path_list:
+    new_file = path_list[path_list.index("-f") + 1:]
```
olena-hul, line 31: You have already defined new_file above, and there is no need to redefine it, because the file name will always go after `-f` flag. So you can create a variable just once and don't change it

olena-hul (changes requested):

olena-hul (approved):

PR #382, opened 2023-06-02
Develop

kucherenkodmytro:


app/create_file.py
```diff
@@ -1 +1,41 @@
...
+def create_directory(path: str) -> None:
+    os.makedirs(path)
+    print(f"Directory created: {path}")
+
+
+def create_file(file_name: str) -> None:
+    if os.path.exists(file_name):
+        mode = "a"
+    else:
+        mode = "w"
```
ArsenPidhoretskyi, line 31: write this inline
```suggestion
    mode = "a" if os.path.exists(file_name) else "w"
```

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+import os
+import sys
+from datetime import datetime
+
+
+def main() -> None:
+    args = sys.argv
+    if "-d" in args:
+        dir_index = args.index("-d")
+        directory_path = os.path.join(*args[dir_index + 1:])
```
ArsenPidhoretskyi, line 10: not true because if -f flag will be after -d:

ArsenPidhoretskyi (changes requested):

vsmutok (changes requested):
1) bad reflection of the year
2) numbers must not start with zero
3) a blank line is missing
4) if i run `python create_file.py -d dir1 dir2 -f file.txt ` twice i get an error

image

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+from datetime import datetime
+
+
+def main() -> None:
+    args = sys.argv
+    if "-d" in args and "-f" not in args:
+        dir_index = args.index("-d")
+        directory_path = os.path.join(*args[dir_index + 1:])
+        if not os.path.exists(directory_path):
+            create_directory(directory_path)
```
Grayder0152, line 12: You can just use `os.makedirs(directory_path, exist_ok=True)`

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+    if "-f" in args and "-d" in args:
+        file_index = args.index("-f")
+        file_name = args[file_index + 1]
+        dir_index = args.index("-d")
+        directory_path = os.path.join(*args[dir_index + 1:file_index])
+        if not os.path.exists(directory_path):
+            create_directory(directory_path)
```
Grayder0152, line 19: the same as above

app/create_file.py
Grayder0152: pls resolve comment

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    with open(os.path.join(directory_path, file_name), mode) as file:
+        file.write(timestamp + "\n")
+        num = 1
+        while True:
+            content = input("Enter content line: ")
+            if content.lower() == "stop":
+                file.write(" \n")
+                break
+            file.write(str(num) + " " + content + "\n")
```
Grayder0152, line 46: use f-string in this case

Grayder0152 (changes requested):

vasylhnatiuk (approved):

vasylhnatiuk (approved):

PR #386, opened 2023-06-08
solution

sbilikepy:


app/create_file.py
```diff
@@ -1 +1,50 @@
...
+import sys
+import datetime
+
+
+def input_parser(flow: list) -> None:
+    path = os.getcwd()
+
+    if "-d" in flow and "-f" in flow:
+        for folder in flow[flow.index("-d") + 1:flow.index("-f"):]:
+            path = os.path.join(path, folder)
```
ArsenPidhoretskyi, line 11: No need to do it with loop. You can also do:
os.path.join(path, *list_with_str)

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #391, opened 2023-06-15
Solution

diana-shyrokikh:


app/create_file.py
```diff
@@ -1 +1,45 @@
...
+from os import makedirs, path
+
+
+def create_file() -> None:
+    try:
+        new_file_name = argv.pop(argv.index("-f") + 1)
+
+        argv.pop(argv.index("-f"))
+    except (IndexError, ValueError):
+        new_file_name = "new_file.txt"
```
oleksiikolii, line 12: You don't need to create file inside this directory if only `-d` flag was passed
"If only -d flag passed, means all items after this flag are parts of the path.
python create_file.py -d dir1 dir2 - creates directory dir1/dir2 inside current directory."

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+
+    path_new_file = path.join(path_new_file, new_file_name)
+    line_count = 0
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    content = (
+        f"\n{timestamp}\n" if path.exists(path_new_file)
+        else f"{timestamp}\n"
+    )
+
+    while True:
```
oleksiikolii, line 31: I think you need to put different processes in separate functions, you do not repeat yourself here but i think it applies still.
"Create separate functions in order not to have repeating blocks of code." <- from checklist

oleksiikolii (changes requested):

oleksiikolii (approved):
GJ!

YuriiKindrat (approved):

PR #393, opened 2023-06-18
Solution

Andriy-Sydorenko:


app/create_file.py
```diff
@@ -1 +1,60 @@
...
+        while True:
+            line = input("Enter content line: ")
+            if line in ["stop", "STOP", "Stop"]:
+                break
+            file.write(f"{line_number} {line}\n")
+            line_number += 1
+        file.write("\n")
+
+
+def main() -> None:
```
Kirontiko, line 23: I think u have to change name of def to "create_file" or something like that

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+        work_with_file(file_path)
+
+    except (IndexError, TypeError, ValueError, FileNotFoundError) as e:
+        print(f"{type(e).__name__}: {e}")
+    except Exception as e:
+        print(f"{type(e).__name__}: {e}")
+
+
+if __name__ == "__main__":
+    main()
```
Kirontiko, line 60: same here

Kirontiko (approved):

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+def work_with_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        timestamp = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(timestamp + "\n")
+
+        print("Enter content for new_file (Enter 'stop' to finish):")
+        line_number = 1
+        while True:
+            line = input("Enter content line: ")
+            if line in ["stop", "STOP", "Stop"]:
```
YuriiKindrat, line 16: could u make like this 
`if line.lower() == "stop"` to avoid creating unnecessary list )

Andriy-Sydorenko: that's dumb mistake))

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+        flags = terminal_input[1:]
+
+        if "-d" in flags:
+            directories = []
+            for argument in range(flags.index("-d") + 1, len(flags)):
+                if flags[argument] == "-f":
+                    break
+                directories.append(flags[argument])
+            directory = path.join(*directories)
+            makedirs(directory, exist_ok=True)
```
YuriiKindrat, line 42: will this work if we will swap a flags ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        while True:
+            line = input("Enter content line: ")
+            if line.lower() == "stop":
+                break
+            file.write(f"{line_number} {line}\n")
+            line_number += 1
+        file.write("\n")
+
+
+def main() -> None:
```
LiudmylaKulinchenko, line 23: Consider split into functions

app/create_file.py
```diff
@@ -1 +1,59 @@
-# write your code here
+import datetime
+import sys
+
+from os import path, makedirs
+
+
+def work_with_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
```
LiudmylaKulinchenko, line 8: Remove from the context manager code that doesn't involve work with a file
It is a best practice and will make the code more clean

LiudmylaKulinchenko (changes requested):

PR #394, opened 2023-06-19
'Solution'

ant-komarov:


Tarasidze (approved):
In my opinion, using "argparse" makes your code more flexible and reliable. 
However, your code is well-structured and logical

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    if "-d" in command:
+        f_index = None
+        for idx, item in enumerate(command):
+            if item == "-f":
+                f_index = idx
+        dirs = command[1: f_index]
+        dirs_path = os.path.join(*dirs)
+        os.makedirs(dirs_path, exist_ok=True)
+        if "-f" not in command:
+            return
```
YuriiKindrat, line 38: could u use backward condition here for 40-41 line ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+        for idx, item in enumerate(command):
+            if item == "-f":
+                f_index = idx
+        dirs = command[1: f_index]
+        dirs_path = os.path.join(*dirs)
+        os.makedirs(dirs_path, exist_ok=True)
+
+    if "-f" in command:
+        file_path = os.path.join(dirs_path, file_name)
+        write_to_file(file_path)
```
YuriiKindrat, line 40: what if we will swap the flags?
will this work ?)

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+entered_command = sys.argv[1:]
+
+
+def write_to_file(path_to_file: str) -> None:
+    current_date = datetime.datetime.now()
+    content = str(current_date.strftime("%Y-%m-%d %H:%M:%S")) + "\n"
+    user_input = ""
+    line_number = 1
+    while user_input != "stop":
+        user_input = input("Enter content line: ")
```
LiudmylaKulinchenko, line 15: You can move this line before while loop to avoid the next condition `if user_input != "stop":`

LiudmylaKulinchenko (changes requested):

app/create_file.py
ant-komarov: Actually I don't understand how I can move input outside the loop, if I need new input everytime?

LiudmylaKulinchenko: Sorry, you can duplicate the same line outside the loop
You can do `user_input = input("Enter content line: ")` insead of `user_input = ""` (which is in the 12 line)

LiudmylaKulinchenko (approved):
Good job! 🎈

PR #396, opened 2023-06-19
Solution

DmytroTsehelnyk:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+                break
+            file.write(f"{line_number} {line}\n")
+            line_number += 1
+
+    if ("-f" in command) and ("-d" in command):
+        f_index = command.index("-f")
+        file_name = command[-1]
+        directory = command[1:f_index]
+        create_directory(directory)
+        with open(path.join(*directory, file_name), "w") as file:
```
diana-shyrokikh, line 30: According to README you should use the "a" mode to append new content if file exists:
"App should add current timestamp at the top and number lines. If file.txt already exists it should add content below"

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+            write_into_file()
+
+    elif "-f" in command:
+        file_name = command[-1]
+        with open(file_name, "w") as file:
```
diana-shyrokikh, line 35: Here is the same problem

diana-shyrokikh (changes requested):

diana-shyrokikh (changes requested):

diana-shyrokikh (approved):
Good Job!

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+            line_number += 1
+        file.write("\n")
+
+    if ("-f" in command) and ("-d" in command):
+        f_index = command.index("-f")
+        file_name = command[-1]
+        directory = command[1:f_index]
+        create_directory(directory)
+        with open(path.join(*directory, file_name), "a") as file:
+            write_into_file()
```
YuriiKindrat, line 32: what if we'll swap a flags ?
try to use argparse for this

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+            line_number += 1
+        file.write("\n")
+
+    if ("-f" in command) and ("-d" in command):
+        file_name = argv.pop(argv.index("-f") + 1)
+        argv.pop(argv.index("-f"))
+        directory = argv[argv.index("-d") + 1:]
+
+        create_directory(directory)
+        with open(path.join(*directory, file_name), "a") as file:
```
LiudmylaKulinchenko, line 32: Better to move the context manager inside write_into_file function. By that you will not include in the context manager the code that doesn't involves working with a file
Also it will help you to store all logic in one function

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+
+def create_file(command: list) -> None:
+    def create_directory(directories: list) -> None:
+        dir_path = path.join(*directories)
+        makedirs(dir_path, exist_ok=True)
+
+    def write_into_file() -> None:
+        current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{current_time}\n")
```
LiudmylaKulinchenko, line 13: You can create a new variable to store the text and then write it all down together to use context manager only in with the code where you work with a file

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (changes requested):
All tests should pass

app/create_file.py
```diff
@@ -1 +1,47 @@
-# write your code here
+from os import path, makedirs
+from datetime import datetime
+from sys import argv
+
+
+def create_file(command: list) -> None:
+    def create_directory(directories: list) -> None:
+        dir_path = path.join(*directories)
+        makedirs(dir_path, exist_ok=True)
```
YuriiKindrat, line 9: no need to create additional variable here)

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+            file.write(f"{content}\n")
+
+    if ("-f" in command) and ("-d" in command):
+        file_name = argv.pop(argv.index("-f") + 1)
+        argv.pop(argv.index("-f"))
+        directory = argv[argv.index("-d") + 1:]
+
+        create_directory(directory)
+        dir_path = path.join(*directory, file_name)
+        write_into_file(dir_path)
```
YuriiKindrat, line 35: same is here

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+
+    elif "-f" in command:
+        file_name = argv.pop(argv.index("-f") + 1)
+        write_into_file(file_name)
+
+    elif "-d" in command:
+        directory = argv[argv.index("-d") + 1:]
+        create_directory(directory)
```
YuriiKindrat, line 43: same )

YuriiKindrat (approved):

PR #400, opened 2023-06-19
solution

YKonoplyov:


IhorVoskoboinikov (approved):
You can use:
path = "" if os.path.exists(path) else os.makedirs(path)
and 
I think it's better to write data to the file once, rather than write it line by line. At the beginning, save the data to a variable and then make one record.

app/create_file.py
```diff
@@ -1 +1,53 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def command_realization() -> None:
+    command = sys.argv[1:]
+    print(command)
```
YuriiKindrat, line 8: remove unnecessary print here)

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+
+    if "-d" in command and "-f" in command:
+        command_directory, *path, command_file, file_name = command
+        create_dir_and_file(path, file_name)
```
YuriiKindrat, line 12: what if we will swap a flags ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+    elif "-f" in command:
+        command_file, file_name = command
+        create_file(file_name)
+
+
+def create_directories(path: tuple) -> str:
+    path = os.path.join(*path)
+
+    if not os.path.exists(path):
+        os.makedirs(path)
```
LiudmylaKulinchenko, line 29: u can use makedirs argument "exist_ok" here and avoid `if not os.path.exists(path):` condition

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌺☀

PR #401, opened 2023-06-19
My solution

kovaliskoveronika:


app/create_file.py
```diff
@@ -1 +1,47 @@
...
+    dirs = command[command.index("-d") + 1:]
+    path = dirs[0]
+    for direct in dirs[1:]:
+        path += f"/{direct}"
+    create_dir(path)
+if "-f" in command and "-d" in command:
+    dirs = command[command.index("-d") + 1: command.index("-f")]
+    path = dirs[0]
+    for direct in dirs[1:]:
+        path += f"/{direct}"
```
OlhaTryhub, line 43: it's better to use os.path.join to make a path. It makes your app crossplatform and you won't need loop, just unpacking

OlhaTryhub: Please, make this changes in the same places of code ☘

OlhaTryhub (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+if "-f" in command and "-d" not in command:
+    print(command[command.index("-f")])
+    create_file(command[command.index("-f") + 1])
+
+if "-d" in command and "-f" not in command:
+    dirs = command[command.index("-d") + 1:]
+    path = dirs[0]
+    for direct in dirs[1:]:
+        path += f"/{direct}"
+    create_dir(path)
```
YevheniiIovchu, line 38: Make path through os.path.join to make this code cross-platform as Olha said and make it in 1 string 34-38 lines

YevheniiIovchu (changes requested):

Oleg-s-s (changes requested):
Split your whole code into functions to make it more modular

OlhaTryhub (approved):

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+        enter = input("Enter content line: ")
+        if enter != "stop":
+            data += f"{index} {enter} \n"
+            index += 1
+    with open(file_name, "a") as file:
+        now = datetime.now()
+        file.writelines(now.strftime("%Y-%m-%d %H:%M:%S"))
+        file.writelines("\n")
+        file.writelines(data)
+        file.writelines("\n")
```
LiudmylaKulinchenko, line 24: Make it DRY

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+        else:
+            create_file(command.file)
+
+    if command.directory:
+        path = os.path.join(*command.directory)
+        create_dir(path)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
LiudmylaKulinchenko, line 45: Moving the code from the `if __name__ == "__main__"` block into functions can be a good approach, especially if you want to reuse the functionality in other parts of your code or if you plan to write tests for the functions

LiudmylaKulinchenko (changes requested):

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+    parser.add_argument("-d",
+                        "--directory",
+                        nargs="+",
+                        help="Create a directory")
+
+    args = parser.parse_args()
+    working_with_command(args)
+
+
+read_command()
```
LiudmylaKulinchenko, line 56: This line should stay in 'if __name__ == "__main__"' if you don't want it to be executed every time you import the module

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+        enter = input("Enter content line: ")
+        if enter == "stop":
+            data += "\n"
+            break
+        data += f"{index} {enter} \n"
+        index += 1
+    with open(file_name, "a") as file:
+        now = datetime.now()
+        file.writelines(now.strftime("%Y-%m-%d %H:%M:%S"))
+        file.writelines(data)
```
LiudmylaKulinchenko, line 23: You can make it in one line

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 😎🚀

PR #402, opened 2023-06-19
solution

mariiahorbova:


erikagayan (approved):

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+                file.write(line + "\n")
+
+    print(f"File '{filename}' created successfully at '{file_path}'.")
+
+
+def process_arguments(args: str) -> None:
+    if "-d" in args:
+        directory_index = args.index("-d") + 1
+        directory_path = os.path.join(*args[directory_index:])
+        os.makedirs(directory_path, exist_ok=True)
```
Oleg-s-s, line 30: image
this part work wrong

Oleg-s-s (changes requested):

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+
+def process_arguments(args: List[str]) -> None:
+    if "-d" in args:
+        directory_index = args.index("-d") + 1
+        path = [*args[directory_index:]]
+        if "-f" in path:
+            p_index = path.index("-f")
+            path = path[:p_index]
+            directory_path = os.path.join(*path)
+        else:
```
LiudmylaKulinchenko, line 34: try to rewrirte without else here and in 37 line. Your code will be more clear and shorter

LiudmylaKulinchenko (changes requested):

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+def process_arguments(args: List[str]) -> None:
+    directory_path = "."
+    if "-d" in args:
+        directory_index = args.index("-d") + 1
+        path = [*args[directory_index:]]
+        if "-f" in path:
+            p_index = path.index("-f")
+            path = path[:p_index]
+        directory_path = os.path.join(*path)
+        os.makedirs(directory_path, exist_ok=True)
```
YuriiKindrat, line 35: will this work if we'll swap the flags
f.e. `-f file.txt -d dir1/dir2`
hint: try to use argparse here, to optimize ur code

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+        with open(file_path, "a") as file:
+            file.write("\n\n")
+            file.write(timestamp + "\n")
+            for line in content:
+                file.write(line + "\n")
+    else:
+        with open(file_path, "w") as file:
+            file.write(timestamp + "\n")
+            for line in content:
+                file.write(line + "\n")
```
YuriiKindrat, line 21: u can use `a` mode for both cases
also make ur code DRY

YuriiKindrat (changes requested):

app/create_file.py
mariiahorbova: image
it works but i get why i need to change

LiudmylaKulinchenko (approved):
Well done 😸

PR #403, opened 2023-06-19
Solution

erikagayan:


app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_file(directory: str, filename: str) -> None:
```
Altahoma, line 6: For what is `directory: str` here if you don't use it?
This function does not create a file by path.
Command: `python create_file.py -d dir1 dir2 -f file.txt` creates the following folders: `dir1 -> dir2 -> -f -> file.txt`, even "file.txt" is a folder and creates the file `file.txt` in the local folder.

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+    content = []
+
+    if os.path.exists(filename):
+        with open(filename, "r") as file:
+            content = file.readlines()
```
Altahoma, line 11: Why do you copy the content if you can just write new lines to the end of the file using mode `a`?

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        new_line = input("Enter content line: ")
+
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    content.insert(0, timestamp + "\n")
+
+    # Write content to file
+    with open(filename, "w") as file:
+        file.writelines(content)
+
+    print("File created successfully.")
```
Altahoma, line 27: There is no need for `print()`.

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+
+def create_directory(directory: str) -> None:
+    os.makedirs(directory)
+    print("Directory created successfully.")
+
+
+def file() -> None:
```
Altahoma, line 35: Better to use descriptive names that describe the function's purpose. For exmaple: `run_app`, `parse_commsnds`, `check_flags`.

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+    if "-d" in sys.argv:
+        directory_index = sys.argv.index("-d") + 1
+        directory = os.path.join(*sys.argv[directory_index:])
+        create_directory(directory)
+
+    if "-f" in sys.argv:
+        filename_index = sys.argv.index("-f") + 1
+        filename = sys.argv[filename_index]
+        create_file(directory, filename)
```
Altahoma, line 44: This will raise an error if we only want to create a file. It also does not provide the ability to write to an existing file.
Check: `python create_file.py -f file.txt`

Altahoma (changes requested):

app/create_file.py
```diff
@@ -37,12 +37,13 @@ def create_file() -> None:
 
     if filename:
         filename = os.path.join(*dir_path, filename)
-        file_time = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n"
+        time = datetime.datetime
```
Altahoma, line 40: No need to write `datetime.datetime` if you import a single module.

Altahoma (changes requested):

Altahoma (approved):
Greate job!

YuriiKindrat (approved):

PR #406, opened 2023-06-19
Solution

YevheniiIovchu:


maccarets (approved):

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+            break
+        line += 1
+        content.append(str(line) + " " + string)
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+    return timestamp + "\n" + "\n".join(content) + "\n\n"
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
LiudmylaKulinchenko, line 26: Will it work correctly if I would like import the logic from this file?
Remove code from `if __name__ == "__main__":` part. Create functions outside it

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):

PR #407, opened 2023-06-19
Solution

imelnyk007:


app/create_file.py
LiudmylaKulinchenko, line 1: Separate your code into functions to make it more modular

LiudmylaKulinchenko (changes requested):

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+import sys
+from datetime import datetime
+
+
+def create_file() -> None:
+    command = sys.argv
+    if "-f" in command:
+        file_name = command[-1]
+        if "-d" in command:
+            dir_list = command[command.index("-d") + 1 : command.index("-f")]
```
LiudmylaKulinchenko, line 11: will this work if we will swap a flags ?
like -f file.txt -d dir1/dir2

LiudmylaKulinchenko (changes requested):

PR #408, opened 2023-06-19
Solution

Po-SvinSkipodla:
I don't have any test. I'm confused with this task, so i need help(

### DIFF
--- a/app/create_file.py
+++ b/app/create_file.py
@@ -1,1 +1,34 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_file(directory, filename: str) -> None:
+    if directory:
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+        path = os.path.join(directory, filename)
+    counter = 1
+    if filename:
+        with open(f"{path}.txt", "a") as new_file:
+            new_file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+            new_line = input("Enter content line: ")
+            while new_line != "stop":
+                new_file.write(f"{counter} {new_line}\n")
+                counter += 1
+                new_line = input("Enter content line: ")
+
+
+def main():
+    directory = ""
+    filename = ""
+    if "-d" in sys.argv:
+        directory = "/".join(sys.argv[sys.argv.index("-d") + 1:])
+    if "-f" in sys.argv:
+        filename = sys.argv[sys.argv.index("-f") + 1:]
+    if directory and filename:
+        directory = "/".join(sys.argv[sys.argv.index("-d") + 1:sys.argv.index("-f") - 1])
+        create_file(directory, filename)
+    create_file(directory, filename)
+
+


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+
+def main() -> None:
+    directory = ""
+    filename = ""
+    if "-d" in sys.argv:
+        directory = "/".join(sys.argv[sys.argv.index("-d") + 1:])
+    if "-f" in sys.argv:
+        filename = sys.argv[sys.argv.index("-f") + 1:]
+    if directory and filename:
+        directory = ("/".join(sys.argv[sys.argv.index("-d")
```
thorium-zero, line 30: You should, probably, use `os.path.join` instead of `str.join`.
Also, you could extract your `-d` and `-f` sections from your arguments prior to join operations to keep your code clean.
On the other hand, it is, probably, better to use the `argparse` here anyway

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+                                       + 1:sys.argv.index("-f") - 1]))
+        create_file(directory, filename)
+    create_file(directory, filename)
```
thorium-zero, line 33: not sure about the line 32, why do you even do it here? Line 33 call will handle this alright

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+        path = os.path.join(directory, filename)
+    counter = 1
+    if filename:
+        with open(f"{path}.txt", "a") as new_file:
+            new_file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+            new_line = input("Enter content line: ")
+            while new_line != "stop":
+                new_file.write(f"{counter} {new_line}\n")
+                counter += 1
+                new_line = input("Enter content line: ")
```
thorium-zero, line 19: the fact that you're repeating yourself with those lines 15 and 19 seem a bit weird. Why wouldn't you initialize your `new_line` variable with None and place your line 19 right after the loop's signature?

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+
+
+def create_file(directory: str, filename: str) -> None:
+    if directory:
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+        path = os.path.join(directory, filename)
+    counter = 1
+    if filename:
+        with open(f"{path}.txt", "a") as new_file:
```
thorium-zero, line 13: I don't like the idea of hardcoding an extension here. What if your arguments will include the extension, and what if that extension should be different?

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+
+
+def main() -> None:
+    directory = ""
+    filename = ""
+    if "-d" in sys.argv:
+        directory = "/".join(sys.argv[sys.argv.index("-d") + 1:])
+    if "-f" in sys.argv:
+        filename = sys.argv[sys.argv.index("-f") + 1:]
+    if directory and filename:
```
thorium-zero, line 29: I believe, that you could do whatever this conditional is intended to do inside the previous one. If you have no `-d` argument, your directory will remain empty, so it wont be a problem to join it to the file name anyway

thorium-zero (changes requested):

Oleg-s-s (approved):
:+1:

PR #409, opened 2023-06-20
Implemented an application for creating files and paths

Altahoma:


app/create_file.py
```diff
@@ -1 +1,54 @@
...
+        line = input("Enter content line: ")
+        if line == "stop":
+            break
+        content.append(f"{line_number} {line}")
+        line_number += 1
+    return content
+
+
+def create_file(path: str) -> None:
+    mode = "a" if os.path.exists(path) else "w"
```
YuriiKindrat, line 21: u can always use mode `a` here)

YuriiKindrat (approved):
GJ

app/create_file.py
Altahoma: Indeed, the `a` mode involves creating a file if it doesn't exist.
Simplified the statement)

PR #411, opened 2023-06-20
Solution

volodymyr-komarnyckyi:


app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+def create_directory() -> str:
+    try:
+        dir_file = os.path.join(sys.argv[sys.argv.index("-d") + 1:])
+    except (TypeError, ValueError):
+        dir_file = ""
+    else:
+        os.makedirs(dir_file)
+
+    return dir_file
```
Oleg-s-s, line 27: Will this work with reversed flags? f.e. `-f ... -d ...`?
Also there is more efficient and easier way to parse data using `argparse` have a look on it.
Using try except is superfluous considering that it is possible to make in more efficient and easier way.

Oleg-s-s (changes requested):
Will

Oleg-s-s (approved):
Good job

PR #414, opened 2023-06-20
done

maccarets:


app/create_file.py
```diff
@@ -1 +1,53 @@
...
+        if line == "stop":
+            break
+        lines.append(f"{line_count} {line}")
+        line_count += 1
+
+    return lines
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "w") as file:
```
YuriiKindrat, line 48: script should add to existing info a new info, if file is exist, will it work correctly with `w` mode ?

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+    if dir_path:
+        create_dir(dir_path)
+
+    if filename:
+        file_path = os.path.join(dir_path, filename)
+        create_file(file_path)
+
+
+def create_dir(dir_path: str) -> None:
+    os.mkdir(dir_path)
```
YuriiKindrat, line 31: mkdir will create only one dirrectory, what if we will pass `dir1/dir2/..`

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+            break
+        lines.append(f"{line_count} {line}")
+        line_count += 1
+
+    return lines
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        for line in content():
```
LiudmylaKulinchenko, line 45: Try to avoid a loop here

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        argv.remove("-f")
+        argv.remove(filename)
+
+    if "-d" in argv:
+        index = argv.index("-d")
+        print(argv)
+        dir_path = os.path.join(*argv[index + 1:])
+
+    if dir_path:
+        os.makedirs(dir_path)
```
LiudmylaKulinchenko, line 23: u can use makedirs argument "exist_ok" here

LiudmylaKulinchenko (changes requested):

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+    return lines
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        file.write("\n".join(content()))
+
+
+parse()
```
Oleg-s-s, line 48: Use correct syntax for module that is main file in your code

Oleg-s-s (changes requested):

Oleg-s-s (approved):
:+1:

PR #416, opened 2023-06-21
Solution

Kasatello:


app/create_file.py
```diff
@@ -1 +1,46 @@
-# write your code here
+import os
+import argparse
+from datetime import datetime
+from typing import Any
+
+
+def create_directory(directory: str) -> None:
+    os.makedirs(directory, exist_ok=True)
```
YuriiKindrat, line 8: it seems to me, there is no need to create func for other func ))

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+
+def create_directory_and_file() -> None:
+    args = parse_arguments()
+    if args.directory:
+        directory_path = os.path.join(*args.directory)
+        create_directory(directory_path)
+    if args.file:
+        file_name = args.file
+        create_file(".", file_name)
```
YuriiKindrat, line 42: so it'll always create file in current dir ? 
it should create file inside dirs, if this arg was passed

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+
+
+def create_directory_and_file() -> None:
+    args = parse_arguments()
+    if args.directory:
+        directory_path = os.path.join(*args.directory)
+        os.makedirs(directory_path, exist_ok=True)
+        if args.file:
+            file_name = args.file
+            create_file(directory_path, file_name)
```
Oleg-s-s, line 38: What if I pass only -f flag?

Oleg-s-s (changes requested):

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+import os
+import argparse
+from datetime import datetime
+from typing import Any
+
+
+def create_file(directory: str, filename: str) -> None:
+    file_path = os.path.join(directory, filename)
+
+    with open(file_path, "a") as file:
```
LiudmylaKulinchenko, line 10: Why do need to open a file already here? Next lines doesn't involve a work with file

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🎉😸

PR #417, opened 2023-06-22
Solution

vkrasnovyd:


app/create_file.py
```diff
@@ -1 +1,53 @@
...
+
+    if directory_name:
+        create_directory(directory_name)
+
+    if file_name:
+        create_file(path.join(directory_name, file_name))
+
+
+def create_directory(directory: str) -> None:
+    makedirs(directory)
```
EdAlekseiev, line 27: If a directory exists it will raise an exception. Pass argument `exist_ok = True` to avoid exception

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #430, opened 2023-07-11
Solution

prykhodkom:


YukihiroSM (changes requested):
The script incorrectly works with -d and -f options together: 
When I executed: python create_file.py -d dir3/dir4 -f file1.txt
I got the folders structure: dir3\dir4\-f\file1.txt

When I executed python create-file.py -f file1.txt -d dir1/dir2
The folders were created correctly, but the file was not created.

YukihiroSM (approved):
Looks good now. Thanks for fixing

PR #434, opened 2023-07-15
Solution

Veinmax:


app/create_file.py
```diff
@@ -1 +1,47 @@
...
+    if "-f" in command:
+        index_f = command.index("-f") + 1
+        create_file(os.getcwd(), index_f)
+
+
+def make_dir(directory: str) -> None:
+    try:
+        os.makedirs(directory)
+    except FileExistsError:
+        pass
```
ArsenPidhoretskyi, line 30: Use os.makedirs(directory, exist_ok=True) instead of try-except

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+
+
+def create_directory() -> None:
+    if "-d" in command:
+        index_d = command.index("-d") + 1
+        if "-f" in command:
+            index_f = command.index("-f") + 1
+            directory = (
+                os.path.join(os.getcwd(), *command[index_d:index_f - 1])
+            )
```
ArsenPidhoretskyi, line 15: But if index f will be first?

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+from datetime import datetime
+import sys
+import os
+
+command = sys.argv
+
+
+def create_directory() -> None:
+    index_d = command.index("-d") if "-d" in command else -1
+    index_f = command.index("-f") if "-f" in command else -1
```
ArsenPidhoretskyi, line 10:
```suggestion
    index_d = command.index("-d") if "-d" in command else None
    index_f = command.index("-f") if "-f" in command else None
```
Better to use None instead of some magic numbers

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #435, opened 2023-07-16
Solution create file from terminal

MarkShapka:


app/create_file.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
```
Polyakiv-Andrey, line 5: rewrite this using functions or classes and to test your code use if __name__ == __main__:

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #442, opened 2023-07-26
solution

vkleshko:


app/create_file.py
```diff
@@ -1 +1,56 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory(directory_path: str) -> None:
+    if not os.path.exists(directory_path):
+        os.makedirs(directory_path)
```
ArsenPidhoretskyi, line 8: Just use
```suggestion
     os.makedirs(directory_path, exist_ok=True)
```

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+            date_today = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            existing_file.write("\n" + date_today + "\n")
+            number = 0
+            while True:
+                content = input("Enter content line: ")
+                if content != "stop":
+                    existing_file.write(f"{1 + number} {content}\n")
+                    number += 1
+                else:
+                    break
```
ArsenPidhoretskyi, line 35: Duplicate code. Just mode changed

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+    if "-d" in sys.argv and "-f" not in sys.argv:
+        directory_path = os.path.join(*sys.argv[sys.argv.index("-d") + 1:])
+        create_directory(directory_path)
+
+    if "-f" in sys.argv and "-d" not in sys.argv:
+        file_name = sys.argv[sys.argv.index("-f") + 1]
+        write_to_file(file_name)
+
+    if "-d" in sys.argv and "-f" in sys.argv:
+        directory_path = os.path.join(*sys.argv[sys.argv.index("-d") + 1:])
```
ArsenPidhoretskyi, line 48: But if -f flag will be after -d?

ArsenPidhoretskyi (changes requested):

PR #446, opened 2023-08-03
Solution

tanya-burdenko:


app/create_file.py
```diff
@@ -1 +1,38 @@
...
+            new_line = input("Enter content line: ")
+            if new_line == "stop":
+                new_file.write("\n")
+                break
+            new_file.write(f"{page} {new_line} \n")
+            page += 1
+
+
+if __name__ == "__main__":
+    if sys.argv[1] == "-d" and sys.argv[-2] == "-f":
```
ArsenPidhoretskyi, line 26: It is not true.
Because this must works the same:

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+        directories = sys.argv[2:-2]
+        file_name = sys.argv[-1]
+        create_directories_and_file(directories, file_name)
+    elif sys.argv[1] == "-f" and sys.argv[3] == "-d":
+        directories = sys.argv[4:]
+        file_name = sys.argv[2]
+        create_directories_and_file(directories, file_name)
```
ArsenPidhoretskyi, line 42:
```suggestion
    if "-d" in sys.argv and "-f" in sys.argv:
        file_index = sys.argv.index("-f") + 1
        directories_index = sys.argv.index("-d") + 1
        directories = sys.argv[directories_index: max(file_index - 1, len(sys.argv))]
        create_directories_and_file(directories, file_name)
```
No need to use magic numbers

ArsenPidhoretskyi (changes requested):

app/create_file.py
tanya-burdenko: I didn't know that, thanks

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #447, opened 2023-08-03
Solution

mwellick:


mwellick:
Not good, I found some mistakes

mwellick:
My code doesn't work correctly
Found some mistakes
I'am preparing new solution

mwellick:
I hope this solution will be better :)

app/create_file.py
```diff
@@ -1 +1,45 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+from typing import Any
+
+
+def create_file_with_content(file_path: str) -> Any:
```
EdAlekseiev, line 7:
```suggestion
def create_file_with_content(file_path: str) -> None:
```

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        while True:
+            input_line = input("Enter content line:")
+            if input_line == "stop":
+                file.write(" \n")
+                break
+            file.write(f"{number_of_content} {input_line}\n")
+            number_of_content += 1
+
+
+def main() -> Any:
```
EdAlekseiev, line 23:
```suggestion
def main() -> None:
```

app/create_file.py
```diff
@@ -1 +1,45 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+from typing import Any
+
+
+def create_file_with_content(file_path: str) -> Any:
+    current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    content_w_time = f"{current_time}"
```
EdAlekseiev, line 9: You don't need use f-string here because `current_time` is already string

EdAlekseiev (changes requested):
Good job

app/create_file.py
mwellick: i forgot to change after my previous solution

EdAlekseiev (approved):
Good job

PR #448, opened 2023-08-03
'create_file.py' app

maxlapchuk:


app/create_file.py
```diff
@@ -1 +1,35 @@
...
+            line_number += 1
+
+
+if "-d" in argv and "-f" in argv:
+    create_directories(path.join(*argv[2:-2]))
+    file_input(create_file_path())
+elif "-d" in argv:
+    create_directories(path.join(*argv[2:]))
+elif "-f" in argv:
+    file_input(argv[-1])
```
ArsenPidhoretskyi, line 35: Move this to function and just call function. No need to write raw code

app/create_file.py
```diff
@@ -1 +1,35 @@
...
+            line = input("Enter content line: ")
+            if "stop" == line.rstrip():
+                work_file.write("\n")
+                break
+            work_file.write(f"{line_number} {line} \n")
+            line_number += 1
+
+
+if "-d" in argv and "-f" in argv:
+    create_directories(path.join(*argv[2:-2]))
```
ArsenPidhoretskyi, line 30: Not true. If I will enter
Your code will not work

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #450, opened 2023-08-04
solution

vsafian:


app/create_file.py
```diff
@@ -1 +1,80 @@
...
+if (dirs_on_path in tokens and first_is_name in tokens
+        and command.index(dirs_on_path) < command.index(first_is_name)):
+    path = plots[dirs_on_path](tokens[dirs_on_path])
+    plots[first_is_name](tokens[first_is_name], path)
+
+else:
+    for flag in tokens:
+        plots[flag](tokens[flag])
+
+sys.exit()
```
vsmutok, line 80: Why you need it?

app/create_file.py
```diff
@@ -1 +1,80 @@
...
+}
+
+if (dirs_on_path in tokens and first_is_name in tokens
+        and command.index(dirs_on_path) < command.index(first_is_name)):
+    path = plots[dirs_on_path](tokens[dirs_on_path])
+    plots[first_is_name](tokens[first_is_name], path)
+
+else:
+    for flag in tokens:
+        plots[flag](tokens[flag])
```
vsmutok, line 78: put it to some function

vsmutok (changes requested):
also this failed 
`python create_file.py -f file2.txt -d dir1 dir2 dir3`

app/create_file.py
```diff
@@ -1 +1,77 @@
...
+    if (dirs_on_path in tokens
+            and first_is_name in tokens):
+        path = create_dirs(tokens[dirs_on_path])
+        create_and_write_file(tokens[first_is_name], path)
+    else:
+        for flag in tokens:
+            plots[flag](tokens[flag])
+
+
+run_cmd_commands()
```
vsmutok, line 77: add 
```python 
if __name__ == '__main__':
```

vsmutok (changes requested):

app/create_file.py
```diff
@@ -1 +1,79 @@
...
+        flag_name = flag + " "
+        if count < len(currents):
+            end_line = command.index(currents[count])
+            current_command = command[start_line: end_line - 1]
+            tokens[flag] = current_command.replace(flag_name, "")
+            command = command.replace(current_command, "")
+        else:
+            current_command = command[start_line:].strip()
+            tokens[flag] = current_command.replace(flag_name, "")
+        count += 1
```
ArsenPidhoretskyi, line 23: So hard code.
Simplify it. Due to task you can only have 2 flags. Realize only for this case

ArsenPidhoretskyi (changes requested):

app/create_file.py
vsafian: I agree this solution has more difficult realization than just got it using if blocks but "in the future" this solution give more abilities, when relative customer wants to add more than 2 flags. So, when I read task readme I looked in that key, bacause it on the surface for me. Or I just enjoing when write intricate code.😂😂))).

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+import sys
+
+DIRS, FILE = "-d", "-f"
+
+
+def tokenize_command(command: list) -> dict:
+    global DIRS, FILE
+    tokens, realize_info = {}, 1
+    if DIRS in command:
+        tokens[DIRS] = command[command.index(DIRS) + realize_info:]
```
ArsenPidhoretskyi, line 14: But your code will fail in this case:

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #451, opened 2023-08-07
Develop

ip-ilya:


app/create_file.py
```diff
@@ -1 +1,42 @@
...
+import os
+import datetime
+
+terminal_commands = argv
+
+if "-d" in terminal_commands:
+    if "-f" in terminal_commands:
+        dirs = (
+            terminal_commands[terminal_commands.index("-d") + 1:
+                              terminal_commands.index("-f")]
```
Dimosphen1, line 11: Try to reduce number of calls for index of values, e.g. `terminal_commands.index("-f")`. You can call it once, write result to a variable and later use that variable for slicing and checks.

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+        )
+    else:
+        dirs = (
+            terminal_commands[terminal_commands.index("-d") + 1:]
+        )
+
+    for directory in dirs:
+        if not os.path.exists(directory):
+            os.mkdir(directory)
```
Dimosphen1, line 20: You could join directories and create all of them at once if absent with the use of `os.makedirs`. Rewrite the code with the use of this approach.

app/create_file.py
```diff
@@ -1 +1,42 @@
-# write your code here
+from sys import argv
+import os
+import datetime
+
+terminal_commands = argv
+
+if "-d" in terminal_commands:
+    if "-f" in terminal_commands:
```
Dimosphen1, line 8: Try to avoid this nesting of `if` conditions. Prefer writing multiple conditions in one line and use `elif` over nesting multiple `if` conditions.

Dimosphen1 (changes requested):

PR #452, opened 2023-08-09
Solution

nschcur:


app/create_file.py
```diff
@@ -1 +1,52 @@
...
+    with open(file_path, "a") as file:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{timestamp}\n")
+        for idx, line in enumerate(content_lines, start=1):
+            file.write(f"{idx} {line}\n")
+
+
+def main() -> None:
+    if len(sys.argv) < 3:
+        print("Usage: python create_file.py [-d directory] -f filename")
```
ArsenPidhoretskyi, line 17: Also you can use it as:
python create_file.py -f filename [-d directory]

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+        return
+
+    content_lines = []
+    file_path = ""
+
+    if "-d" in sys.argv:
+        dir_index = sys.argv.index("-d") + 1
+        directory = os.path.join(*sys.argv[dir_index:])
```
ArsenPidhoretskyi, line 25: If command will be:

ArsenPidhoretskyi: Your code will not work

ArsenPidhoretskyi (changes requested):

PR #453, opened 2023-08-09
Solution

romkapomka12:


app/create_file.py
```diff
@@ -1 +1,37 @@
...
+    else:
+        dir_path = ""
+    if "-f" in command:
+        f_index = command.index("-f")
+        file_name = command[f_index + 1]
+        create_file(dir_path + file_name)
+
+
+if __name__ == "__main__":
+    create_file("file.name")
```
YukihiroSM, line 37: Filename should be set by user, not hardcoded.

YukihiroSM (changes requested):
image

YukihiroSM (changes requested):
There should be some entrypoint, because we are going to run the function from terminal. In your previous solution you had something like `if __name__ == "__main__":` Add this and execute some entry function

romkapomka12:
> There should be some entrypoint, because we are going to run the function from terminal. In your previous solution you had something like `if __name__ == "__main__":` Add this and execute some entry function

I changed the code. Now the directory and file are being created

EdAlekseiev (approved):
Good job

PR #463, opened 2023-08-22
Solution

Starchild666:


app/create_file.py
```diff
@@ -1 +1,69 @@
...
+        dirs, file_name = command_data.get("dirs"), command_data.get(
+            "file_name"
+        )
+        file_path = os.path.join(*dirs, file_name)
+
+        directory = os.path.dirname(file_path)
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+            with open(file_path, "w"):
+                pass
```
ArsenPidhoretskyi, line 52:
```suggestion
        os.makedirs(directory, exist_ok=True)
        with open(file_path, "w"):
            pass
```

ArsenPidhoretskyi (changes requested):

app/functions.py
```diff
@@ -0,0 +1,41 @@
+import sys
+from datetime import datetime
+
+
+def validator(command: list) -> object:
+    if (
+        len(command) >= 4
+        and command[0] == "-d"
+        and "-f" in command[2:]
+        and command[-1] != "-f"
```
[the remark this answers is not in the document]
Starchild666: Hardcode because -d also can be not in first place --->  python create_file.py -d dir1 dir2 -f file.txt  :   
With this example in the task, I thought we had a "hard structure" of command when both -d and -f were presented.

Also why you check by 4?:  if both -d and -f in command -> "-d" + at least 1 dir + "-f" + file_name = at least 4 arguments have to be in place.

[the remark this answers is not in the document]
ArsenPidhoretskyi: But:
python create_file.py -d dir1 dir2 dir3 dir4

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+)
+
+command_to_execute = sys.argv[1:] if sys.argv else []
+current_dir = os.getcwd()
+
+
+def create_file() -> None:
+    validation_result = validator(command_to_execute)
+
+    if validation_result == 1:
```
ArsenPidhoretskyi, line 17: 1, 2, 3 magic Number, create and use constants please

app/functions.py
```diff
@@ -0,0 +1,58 @@
...
+    if len(file_content) == 0:
+        with open(file_name, "a") as f:
+            info_to_write = processed_input_data().rstrip("\n")
+            f.write(info_to_write)
+    else:
+        with open(file_name, "a") as f:
+            info_to_write = "\n" * 2 + processed_input_data().rstrip(
+                "\n"
+            )
+            f.write(info_to_write)
```
ArsenPidhoretskyi, line 48: Here is changed only \n * 2
Try to not duplicate code

app/functions.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def dir_and_file_name_extraction(command: list) -> dict:
+    dirs_file_info = {"dirs": [], "file_name": ""}
+
+    index_of_flag = command.index("-f")
+    dirs_file_info["dirs"] = command[1:index_of_flag]
```
ArsenPidhoretskyi, line 55: But if -f flag will be first:

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
.

app/functions.py
Starchild666: 1. python create_file.py -d dir1 dir2 -f file.txt == python create_file.py -f file.txt -d dir1 dir2) -> True?   OR:
2. python create_file.py -f file.txt -d dir1 dir2:   create file in current dir and dir1/dir2 just standby dirs?

vasylhnatiuk (approved):

PR #466, opened 2023-08-23
Solution

Anastasia-Su:


app/create_file.py
```diff
@@ -1 +1,51 @@
...
+
+    args = parser.parse_args()
+
+    directories = args.directories
+    file_name = args.file_name
+
+    if directories:
+        directory_path = os.path.join(*directories)
+        if not os.path.exists(directory_path):
+            os.makedirs(directory_path)
```
ArsenPidhoretskyi, line 23: Instead of if-statement use exist_ok=True

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,50 @@
...
+
+    if directories:
+        directory_path = os.path.join(*directories)
+        os.makedirs(directory_path, exist_ok=True)
+
+    if file_name:
+        file_path = (os.path.join(*directories, file_name)
+                     if directories
+                     else file_name
+                     )
```
vsmutok, line 28: use `black` to fix indentation

vsmutok (changes requested):

app/create_file.py
Anastasia-Su: I used "black", but length limitation issues were spotted by flake8. Fixed them manually. Hope, current indentation is better.

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+            if os.path.getsize(file_path) > 0:
+                add_space = True
+
+            if add_space:
+                file_to_create.write("\n")
+
+            file_to_create.write(current_timestamp + "\n")
+
+            line_number = 0
+            while 1:
```
ArsenPidhoretskyi, line 41:
```suggestion
            while True:
```
Better to write True instead of 1

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+                file_content = input("Enter content line: ")
+                line_number += 1
+                if file_content == "stop":
+                    break
+
+                file_to_create.write(f"{line_number} {file_content}\n")
```
ArsenPidhoretskyi, line 47: Also move this to separate function. Better to have 2 small function, instead of 1 big

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #467, opened 2023-08-25
Solution

VasylTurok:


app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+    if "-f" in argv:
+        file_path = argv[-1]
+        argv = argv[:-2]
+        print(file_path)
+        print(argv)
+    else:
+        raise CommandNotFound(
+            "Usage: python create_file.py -d [dir1 dir2 ...] -f filename.txt"
+        )
```
ArsenPidhoretskyi, line 22: Not true because -d and -f may not exist

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+
+    if "-d" in argv:
+        dir_index = argv.index("-d") + 1
+        print(dir_index)
+        dir_path = os.path.join(*argv[dir_index:])
```
ArsenPidhoretskyi, line 27: But in this case if command will be:
dir_path will be dir1/dir2/-f/file.txt

ArsenPidhoretskyi (changes requested):

app/create_file.py
VasylTurok: won't because in the previous condition I reassigned the argv variable
image

Polyakiv-Andrey (approved):
Good job!

PR #469, opened 2023-08-27
Solution

OlPlaksa:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+
+if "-f" in sys.argv and "-d" not in sys.argv:
+
+    create_file()
+
+
+if "-d" in sys.argv and "-f" in sys.argv:
+
+    create_dir()
+    create_file()
```
ArsenPidhoretskyi, line 44: Why you can not just use two cases:
if -d in argv
if -f in argv

app/create_file.py
```diff
@@ -1 +1,44 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_dir() -> None:
+    if "-f" in sys.argv:
+        dir_path = os.path.join(*sys.argv[2:-2])
```
ArsenPidhoretskyi, line 8: But if command will be:

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+    else:
+        dir_path = os.path.join(*sys.argv[2:])
+    os.makedirs(dir_path, exist_ok=True)
+
+
+def create_file() -> None:
+    if "-d" in sys.argv:
+        file_path = os.path.join(*sys.argv[2:-2]) + "\\" + sys.argv[-1]
```
ArsenPidhoretskyi, line 16:
```suggestion
        file_path = os.path.join(*sys.argv[2:-2], sys.argv[-1])
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #473, opened 2023-08-28
Solution

nyanplague:


RPodolets (approved):

OlegatorLE (changes requested):
App should add current timestamp at the top and number lines. I don't see any number lines.

OlegatorLE (approved):

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+            if content == "stop":
+                break
+            file_pointer.write(f"{line} {content} \n")
+            line += 1
+
+
+if __name__ == "__main__":
+    command_line = sys.argv
+
+    if (
```
Kroll410, line 30: Put all this logic into a function and call it here

Kroll410 (approved):

PR #477, opened 2023-08-29
'Solution'

lgSilay:


app/create_file.py
```diff
@@ -1 +1,62 @@
...
+def create_dir() -> str:
+    if "-f" in sys.argv:
+        path = os.path.join(
+            "/".join(sys.argv[sys.argv.index("-d") + 1: sys.argv.index("-f")])
+        )
+        if not os.path.exists(path):
+            os.makedirs(path)
+        return path
+
+    path = os.path.join("/".join(sys.argv[sys.argv.index("-d") + 1 :]))
```
ORENYT, line 15: Use os.path.join(current_path, folder) instead of using / or \

ORENYT (changes requested):
Won't work on different OS

ORENYT (approved):

WonnaKoo (approved):

WonnaKoo (approved):

PR #479, opened 2023-08-29
Solution

Illia-Kononenko:


MarianKovalyshyn (approved):
I think you could use better (longer) names for variables but everything else looks good

PR #481, opened 2023-08-29
Implemented app that creates directories and file

MarianKovalyshyn:


CaCtUs1k (approved):

Kroll410 (changes requested):
Use functions, overally looks ok

PR #496, opened 2023-09-02
Test(py-create-file-from-terminal)

baranovr:


app/create_file.py
```diff
@@ -1 +1,54 @@
-# write your code here
+import os
+import sys
+import datetime
+from typing import Any
+
+
+def create_directory(path_parts: Any) -> None:
```
Bohdan-Salamakha, line 7: specify typehint

Bohdan-Salamakha (changes requested):
file.txt should be a file but not a folder
image

PR #500, opened 2023-09-05
'Solution'

WonnaKoo:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+            data = input("Enter content line: ")
+            if data == "stop":
+                file.write("\n")
+                break
+            file.write(f"{number_line} {data} \n")
+            number_line += 1
+
+
+def main() -> None:
+    if "-d" in sys.argv and "-f" in sys.argv:
```
9rosLove, line 28: You could use argparse instead

9rosLove (approved):

PR #510, opened 2023-09-14
Solution

AllaKuksa:


app/create_file.py
```diff
@@ -1 +1,49 @@
...
+def write_content_to_file(file_name: Any) -> None:
+
+    while True:
+        line = input("Enter content line: ")
+        if line == "stop":
+            break
+        file_name.write(line + "\n")
+
+
+commands = sys.argv
```
ArsenPidhoretskyi, line 16: It is so bad to write raw code in center of file

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+
+
+def create_path(directory_path: list) -> None:
+    if "-d" in commands:
+        directory_path = commands[commands.index("-d") + 1:]
+        path = os.path.join(*directory_path)
+
+        if not os.path.exists(path):
+            os.makedirs(path)
```
ArsenPidhoretskyi, line 25: Use exist_ok=True flag instead of check

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+            write_content_to_file(file)
+
+        if os.path.exists(file_name):
+            with open(file_name, "a") as exists_file:
+                exists_file.write(f"\n{file_data}\n")
+                write_content_to_file(exists_file)
+
+
+def create_file_and_path() -> None:
+    if "-d" in commands and "-f" in commands:
```
ArsenPidhoretskyi, line 45: But if only one flag present?

ArsenPidhoretskyi (changes requested):
Your code will do nothing because main function is not called

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+        if line == "stop":
+            break
+        content += f"{num_of_line} {line}\n"
+    content += "\n"
+
+    with open(file_name, "a") as file:
+        file.write(content)
+
+
+def create_file(command: list) -> None:
```
vasylhnatiuk, line 22:
```suggestion
def create_file(command: list[str]) -> None:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #515, opened 2023-09-18
'Solution'

usKos:


app/create_file.py
```diff
@@ -1 +1,69 @@
...
+    if sys.argv.count("-f") and sys.argv.count("-d") == 0:
+        create_file(list_avrg)
+
+    if sys.argv.count("-d") and sys.argv.count("-f"):
+        path_dir = list_avrg[:list_avrg.index("-f")]
+        create_dir(path_dir)
+        create_file(list_avrg)
+
+
+funct_create_file_from_terminal()
```
YukihiroSM, line 69: It's better to do it using `if __name__ == "__main__":`

YukihiroSM (approved):
The code is good, but take a look into comments.

PR #521, opened 2023-09-27
Solution

VitaliyZhabskyi:


vsmutok (changes requested):
* `python create_file.py -d dir1 dir2 -f file.txt` - not create  dir2
* if I have dir1 then I try `python create_file.py -d dir1 dir2` - `dir2` not created
* add line number 
image

app/create_file.py
```diff
@@ -1 +1,48 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+def create_directory(directory: any) -> None:
```
Yurnerroo, line 6: replace `any` with specific type

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        file.write(current_time + "\n")
+        for i, line in enumerate(content):
+            file.write(f"{i + 1} {line}\n")
+
+
+def main() -> None:
+    if len(sys.argv) < 5:
+        print("Incorrect format. Please use -d and -f "
+              "flags to create a directory and a file with content.")
+        return
```
Yurnerroo, line 31: redundant return

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+    if "-d" not in sys.argv or "-f" not in sys.argv:
+        print("Both -d and -f flags are "
+              "required to create a directory and a file.")
+        return
```
Yurnerroo, line 36: same here

Yurnerroo (changes requested):

vsmutok (changes requested):
have you tested your code????

Screenshot 2023-10-02 at 20 59 22
1. dir2, dir3 not created 
2. data should be appended to the file, not overwritten

vasylhnatiuk (approved):
Знімок екрана 2023-10-03 о 21 07 54
your function should work only with -f or -d

PR #523, opened 2023-09-30
solution

antonnech2309:


app/create_file.py
```diff
@@ -1 +1,40 @@
...
+            if line == "stop":
+                break
+            file.write(f"{linestart} {line}\n")
+            linestart += 1
+
+
+argv = sys.argv
+
+if "-d" in argv and "-f" in argv:
+    directories = argv[argv.index("-d") + 1: argv.index("-f")]
```
pavlejviki, line 31: if the user changes the order of the arguments (-f comes before -d: python create_file.py -f file.txt -d dir1 dir2), the program won't behave as intended. It should be able to handle arguments in any order.

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+            line = input("Enter content line:")
+            if line == "stop":
+                break
+            file.write(f"{linestart} {line}\n")
+            linestart += 1
+
+
+argv = sys.argv
+
+if "-d" in argv and "-f" in argv:
```
pavlejviki, line 30: Place the script's logic into a separate function.

pavlejviki (changes requested):

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+        path = make_directory(directories)
+        make_file(filename, path)
+    elif "-d" in argv:
+        path = argv[argv.index("-d") + 1:]
+        make_directory(path)
+    elif "-f" in argv:
+        filename = argv[argv.index("-f") + 1]
+        make_file(filename)
+
+
```
pavlejviki, line 45: Add 
 if __name__ == "__main__":
 
so if you import this script into another script using import, it won't automatically execute processing_arguments(), giving the importing script control over when and how to use the functions within.

pavlejviki (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+import sys
+import os
+
+
+def make_directory(dir_list: list) -> str:
+    path = ""
+    for directory in dir_list:
+        path = os.path.join(path, directory)
+        if not os.path.exists(path):
+            os.mkdir(path)
```
viktoria-rybenchuk, line 12: Create directory with exist_ok=True

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #524, opened 2023-09-30
solution

Shum430:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+    directory = ""
+    filename = ""
+
+    if "-d" in sys.argv:
+        d_index = sys.argv.index("-d")
+        directory = os.path.join(
+            *sys.argv[d_index + 1:
+                      d_index + 1
+                      + sys.argv[d_index + 1:].index("-f")])
```
pavlejviki, line 42: if the user changes the order of the arguments (-f comes before -d: python create_file.py -f file.txt -d dir1 dir2), the program won't behave as intended. It should be able to handle arguments in any order.

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        content.append(line)
+
+    with open(file_path, "a") as file:
+        file.write(timestamp + "\n")
+        for i, line in enumerate(content, start=1):
+            file.write(f"{i} {line}\n")
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
```
pavlejviki, line 30: Place the script's logic into a separate function.

pavlejviki (changes requested):

PR #528, opened 2023-10-04
Solution

sachavskyi:


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+                if words == "stop":
+                    break
+                f.write("\n")
+                f.write(f"{i} {words}")
+                i += 1
+            f.write("\n")
+            f.write("\n")
+
+
+create_file()
```
Dimosphen1, line 33: There are several issues with the code:
- You should not leave two blank lines at the end of the file after the `stop` command. In both cases when you create a new file or append to an existing one, there should not be blank lines at the end.
- You should implement the possibility of running commands with `-d` only, the presence of the `-f` flag should not be obligatory.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #529, opened 2023-10-08
solution

uzlss:


app/create_file.py
```diff
@@ -1 +1,32 @@
...
+    line_args = sys.argv
+    index = 1
+    while index < len(line_args):
+        if line_args[index] == "-d":
+            os.makedirs(os.path.join("app", line_args[index + 1]))
+            os.makedirs(os.path.join("app", line_args[index + 2]))
+            index += 3
+        elif line_args[index] == "-f":
+            create_file(os.path.join("app", line_args[index + 1]))
+            index += 2
```
ArsenPidhoretskyi, line 29: Use .index method for list instead of usage while loop

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+        file.write("\n")
+
+
+def read_line() -> None:
+    line_args = sys.argv
+    index = 1
+    while index < len(line_args):
+        if line_args[index] == "-d":
+            os.makedirs(os.path.join("app", line_args[index + 1]))
+            os.makedirs(os.path.join("app", line_args[index + 2]))
```
ArsenPidhoretskyi, line 25: Your code will not work if dirs will be more that 2, or 1

ArsenPidhoretskyi (changes requested):

vsmutok (changes requested):
have you tested your code?????


1) executing this command `python create_file.py -d dir1 dir2 -f file.txt` a second time will cause an error
2) you need to create a folder in a folder and in the last folder a file
Screenshot 2023-10-09 at 18 12 16

vsmutok (changes requested):
no need to create `app` folder
image

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+import os
+import sys
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        file.write(str(datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+                   + "\n")
+        line = input("Enter content line: ")
+        index = 1
```
vsmutok, line 11: DRY 

here and In 13, 14 lines

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+    if "-f" in line_args:
+        file_name = line_args[line_args.index("-f") + 1]
+        if path:
+            file_path = os.path.join(*path, file_name)
+        else:
+            file_path = os.path.join(file_name)
+        create_file(file_path)
+
+
+read_line()
```
vsmutok, line 42: use `if __name__ == '__main__':`

vsmutok (changes requested):

app/create_file.py
uzlss: is there any change for "index"?
image

uzlss: for loop i can use "while True", but what's about index?

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+import datetime
+import os
+import sys
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        file.write(str(datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+                   + "\n")
+        i = 1
```
vasylhnatiuk, line 10: i is a bad name, rename please

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #537, opened 2023-10-17
solution

MishkaGAMII:


app/create_file.py
```diff
@@ -1 +1,50 @@
...
+
+
+def main() -> None:
+    argv = sys.argv
+    if "-d" in argv and "-f" not in argv:
+        create_dir(argv[argv.index("-d") + 1:])
+
+    elif "-f" in argv and "-d" in argv:
+        path = create_dir(argv[argv.index("-d") + 1:
+                               argv.index("-f")])
```
ArsenPidhoretskyi, line 40: In this case when -f will be first, your core will not work

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+            if record_info.lower() == "stop":
+                file.write("\n\n")
+                break
+            file.write(
+                f"\n{number_line} {record_info}"
+            )
+            number_line += 1
+
+
+def create_dir(path_l : list) -> str:
```
vasylhnatiuk, line 27:
```suggestion
def create_dir(path_l : list[str]) -> str:
```

vasylhnatiuk (changes requested):

PR #541, opened 2023-10-21
sol

Familenko:


app/create_file.py
```diff
@@ -1 +1,35 @@
...
+    dir_location = os.path.join(*dir_locations)
+    os.makedirs(dir_location, exist_ok=True)
+    return dir_location
+
+
+def make_file(file_name: str) -> None:
+    with open(file_name, "a") as file:
+        file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
+        print("Enter content or 'stop'")
+        for counter, line in enumerate(iter(input, "stop"), 1):
```
prochigor, line 16: I think it is better to use while cycle

prochigor (approved):

app/create_file.py
fsocie7y: yep, it could be better

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #542, opened 2023-10-22
solution

SerhiiBudko:


app/create_file.py
```diff
@@ -1 +1,39 @@
...
+
+    elif "-d" in sys.argv:
+        make_dir(sys.argv[sys.argv.index("-d") + 1:])
+
+    elif "-f" in sys.argv:
+        make_file(sys.argv[sys.argv.index("-f") + 1])
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 39: Revise your logic, currently, the interface does not match the one described in the task. After the script starts, in case the `-f` flag is passed, you should wait for the user input and print in the terminal `Enter content line: `. Besides that, you should actually write content to a file, currently script does not perform writing of the user input. Also, pay attention to blank lines, there should be no blank lines at the end of a created or edited file and there should be one blank line between different input sessions, the file content should exactly match the task description.

app/create_file.py
```diff
@@ -1 +1,39 @@
-# write your code here
+import os
+import sys
```
Dimosphen1, line 2: Separate `import ...` and `from ... import ...` with one blank line.

Dimosphen1 (changes requested):

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+def make_directory(directories: list) -> str:
+    directory_path = os.path.join(*directories)
+    os.makedirs(directory_path, exist_ok=True)
+    return directory_path
+
+
+def make_file(file_name: str) -> None:
+    with open(os.path.join(file_name), "a") as file:
+        date_stamp = datetime.now().strftime("%Y-%m-%d %H-%M-%S") + "/n"
```
pavlejviki, line 15: you're using the wrong character for creating a new line. In Python, the character for a new line is \n, not /n.

pavlejviki (changes requested):

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+def make_file(file_name: str) -> None:
+    with open(os.path.join(file_name), "a") as file:
+        date_stamp = datetime.now().strftime("%Y-%m-%d %H-%M-%S") + "\n"
+        file.write(date_stamp)
+        while True:
+            content = input("Enter content line: ") + "\n"
+            if content == "stop\n":
+                break
+            file.write(content)
+        file.write("\n")
```
Dimosphen1, line 22: Your code leaves two blank lines at the end of the file. Change logic to not leave them. Besides that, you should enumerate lines as mentioned in the task description. Also, your code should work in case several flags are passed, e.g. `python create_file.py -d dir1 dir2 -f file.txt`.

Dimosphen1 (changes requested):

app/create_file.py
SerhiiBudko: @Dimosphen1 
I couldn't find where my code left two blank lines at the end. 
In fact, I didn't realise why my code doesn't work with several flags (-d) and (-f). As I know it my be in the def main()?! 

As I understand it, I have enumerated lines in the 'def make_file()' function. I think that is DONE:(

I don't understand how to perform the task, I need some help! I hope that You give me advices during tech-check lesson.

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+
+def make_file(file_name: str) -> None:
+    with open(os.path.join(file_name), "a") as file:
+        date_stamp = datetime.now().strftime("%Y-%m-%d %H-%M-%S") + "\n"
+        file.write(date_stamp)
+
+        line_number = 0
+        while True:
+            line_number += 1
```
viktoria-rybenchuk, line 19: you should add line when you have written content

viktoria-rybenchuk (changes requested):

PR #545, opened 2023-10-22
Solution v0.1

petrykivd:


OlehShumov (approved):

app/create_file.py
```diff
@@ -1 +1,63 @@
...
+
+
+def get_d_index(command: str) -> int:
+    if "-d" in command:
+        return command.index("-d")
+
+
+def get_file_name(command: list) -> str:
+    f_index = get_f_index(command)
+    if f_index is not None:
```
fsocie7y, line 21: if cond: ...   its better than if cond is not None: ...
fix it everywhere

fsocie7y (changes requested):

app/create_file.py
petrykivd: > if cond: ... its better than if cond is not None: ... fix it everywhere

I use it here, because f_index can be 0, and after that in condition i get False.

fsocie7y (approved):
Well done 🚀

PR #548, opened 2023-10-23
Solution

Lyutillis:


app/create_file.py
```diff
@@ -1 +1,36 @@
...
+        f_index -= 2
+    dir_path = "/".join(command[d_index:f_index]) + "/"
+    path = os.path.dirname(dir_path)
+    os.makedirs(path, exist_ok=True)
+    return path
+
+
+
+def write_file(command: list, path: str = "") -> None:
+    filename = path + "/" + command[-1]
```
kostomeister, line 17: its better to use os.path.join(path, command[-1]) because you will never know on which system your code will be ran

kostomeister (approved):
good job, but change your code a bit

app/create_file.py
fsocie7y: yep, use os.path.join()

fsocie7y (changes requested):

app/create_file.py
```diff
@@ -1 +1,35 @@
...
+import sys
+import os
+import datetime
+
+
+def create_dirs(command: list) -> str:
+    d_index, f_index = 2, len(command)
+    if "-f" in command:
+        f_index -= 2
+    dir_path = "/".join(command[d_index:f_index]) + "/"
```
LiudmylaKulinchenko, line 10: You should also fix here

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #550, opened 2023-10-23
Solution

Naz-iv:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+dir_path = None
+
+_, *command = argv
+
+if "-d" in command and "-f" in command:
+    dir_path, file_name = "/".join(command[1:]).split("-f")
+    file_name = path.dirname(dir_path) + file_name
+
+elif "-d" in command:
+    dir_path = "/".join(command[1:])
```
RabenkoYevhenii, line 16: maybe you should use here os.path.join instead of "/".join. It was sad it`s a bad example in checklist.md to use "/".join

RabenkoYevhenii (approved):
It is good to me, but have to correct join

app/create_file.py
fsocie7y: yep, fix it

fsocie7y (changes requested):

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+        output_data.append(f"{comment}{line_number}  {line}\n")
+
+    return output_data
+
+
+def file_handler(file_name: str, data: List[str]) -> None:
+
+    access_mod = "w"
+    if path.exists(file_name):
+        access_mod = "a"
```
LiudmylaKulinchenko, line 52: You can use just `a` mode, file will be created with it

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+
+    with open(file_name, access_mod) as f:
+        f.writelines(data)
```
LiudmylaKulinchenko, line 55: Better to use more descriptive name here

LiudmylaKulinchenko (changes requested):

app/create_file.py
Naz-iv: fixed

app/create_file.py
Naz-iv: fixed

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+from sys import argv
+from os import path, makedirs
+from typing import List
+
+
+def command_reader(command: tuple) -> str:
+    file_name = "file.txt"
+    dir_path = None
+
+    if "-d" in command and "-f" in command:
```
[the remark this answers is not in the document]
Naz-iv: Fixed this using argparser, maybe worth mentioning in task description, that argparser can be used, so we don't come up with manual ways of parsing command.

fsocie7y (approved):
GJ 🚀

PR #554, opened 2023-10-23
Solution

Katherine-Greg:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+def create_directory(file_path: list[str]) -> str:
+    path = os.path.join(*file_path)
+    os.makedirs(path, exist_ok=True)
+    return path
+
+
+def main() -> None:
+    argv = sys.argv
+    if "-d" in argv and "-f" not in argv:
+        filename = argv[argv.index("-f") + 1]
```
SelFisher1488, line 27: `if "-f" not in argv` means that there is no "-f" index

SelFisher1488 (approved):
great code but some logic errors

app/create_file.py
fsocie7y: yep, agree

fsocie7y (changes requested):

LiudmylaKulinchenko (approved):
Good job!

PR #559, opened 2023-10-24
Solution

Wonsky1:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+import sys
+import os
+from datetime import datetime
+from typing import List
+
+
+def create_dirs(dirs: List[str]) -> str:
+    if not os.path.exists("/".join(dirs)):
+        os.makedirs("/".join(dirs))
+    return "/".join(dirs)
```
fsocie7y, line 10: always use os.path.join() for cases like this)

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #567, opened 2023-10-24
Solution

Lebwa1337:


Dimosphen1 (changes requested):

Lebwa1337:
i think i get it all, but i can`t figure out,how i don`t leave one blank line at the end of the file

Polyakiv-Andrey (changes requested):

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+from sys import argv
+
+
+def create_file(path_with_file: str) -> None:
+    with open(os.path.join(path_with_file), "a") as file:
+        content = ""
+        enumerate_line = 1
+        while True:
+            line = str(input("Enter content line: "))
+            if line == "stop":
```
pavlejviki, line 12: The input() function already returns a string by default, so there's no need to use str().

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #569, opened 2023-10-26
Solution

Andrii-Silenko:


Familenko (approved):

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+    with open(file_path, "a") as f:
+        f.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+        while True:
+            user_line = input("Enter content line: ")
+            if user_line == "stop":
+                break
+            f.write(user_line + "\n")
+
+
+if "-d" in sys.argv and "-f" in sys.argv:
```
LiudmylaKulinchenko, line 16: You can rewrite this function with only two checks, one for -d and one for -f, and with way less code duplication.
Better yet, you can also use argparse module here — it would be easier, more readable, and you won't have to reinvent the wheel.

LiudmylaKulinchenko, line 16: Place it in function

PR #570, opened 2023-10-28
Solution

okien1:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        file.write("\n")
+
+
+def create_dirs(dirs_names: list[str]) -> str:
+    if not dirs_names:
+        return ""
+    path = os.path.join(*dirs_names)
+    if not os.path.isdir(path):
+        os.makedirs(path)
+    return path
```
viktoria-rybenchuk, line 25: This code can be replaced with a single call to `os.makedirs with exist_ok=True`

viktoria-rybenchuk (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+        line_input = input("Enter content line:")
+        while line_input != "stop":
+            file.write(f"{count_of_iteration} {line_input}\n")
+            count_of_iteration += 1
+            line_input = input("Enter content line:")
+        file.write("\n")
+
+
+def create_dirs(dirs_names: list[str]) -> str:
+    if not dirs_names:
```
viktoria-rybenchuk, line 20: If 'dirs_names' is not empty, create the corresponding directories and return the path, otherwise no action is taken, and the function returns None.

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+import sys
+import os
+from datetime import datetime
+
+
+def write_content_to_file(path_and_file_name: list[str]) -> None:
+    with open(os.path.join(*path_and_file_name), "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        line_input = 0
+        count_of_iteration = 1
```
viktoria-rybenchuk, line 10: use descriptive variable name as `line_count`

app/create_file.py
```diff
@@ -1 +1,47 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def write_content_to_file(path_and_file_name: list[str]) -> None:
+    with open(os.path.join(*path_and_file_name), "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        line_input = 0
```
viktoria-rybenchuk, line 9: why do you need this `line_input = 0`?

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #575, opened 2023-10-31
Wrote function as requested

sberdianskyi:


app/create_file.py
```diff
@@ -1 +1,38 @@
-# write your code here
+import datetime
+import os
+import sys
+
+
+def create_path(command: list) -> None:
```
Yurnerroo, line 6: specify list annotation

Yurnerroo (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #576, opened 2023-10-31
'Solution_create-file-from-terminal'

vvg04021968:


app/create_file.py
```diff
@@ -1 +1,37 @@
...
+
+elif "-d" in sys.argv and "-f" in sys.argv:
+    d_index = sys.argv.index("-d")
+    f_index = sys.argv.index("-f")
+    directory_name = os.path.join(*sys.argv[d_index + 1:f_index])
+    file_name = sys.argv[-1]
+
+    if not os.path.exists(directory_name):
+        os.makedirs(directory_name)
+    create_file(os.path.join(directory_name, file_name))
```
pavlejviki, line 37: all logic should be inside a main() function

pavlejviki (changes requested):

Dimosphen1 (changes requested):

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+        directory_name = os.path.join(*sys.argv[d_index + 1:f_index])
+        file_name = sys.argv[-1]
+
+        if not os.path.exists(directory_name):
+            os.makedirs(directory_name)
+        create_file(os.path.join(directory_name, file_name))
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 42: There are several issues with the current logic:

- You should enumerate your lines as it is shown in the example;
- You should not make two blank lines at the end of the newly created file or when you append to an existing one. After the two iterations of command the file content should be the same as it is mentioned in the task description:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content
```

Dimosphen1 (changes requested):

PR #577, opened 2023-10-31
Solution

loplicat:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+    elif "-f" in input_data:
+        file_name = input_data[input_data.index("-f") + 1]
+        return create_file(file_name)
+    elif "-d" in input_data:
+        dirs = input_data[input_data.index("-d") + 1:]
+        return create_path(dirs)
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 48: There are several issues with your solution:

- Your code does not enumerate lines as it should be done according to the example;
- You do not make one blank line between the previous input and a new input, when you operate on the same file for several times. Pay attention that you should not leave a blank line at the end of the file. Your file should be as following after running two sessions on the same file:

```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content
``` 

Resolve these issues upon the next review request.

Dimosphen1 (changes requested):

app/create_file.py
loplicat: Thanks, I've just implemented this changes.

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+import os
+
+
+def create_path(directories: list) -> str:
+    path = os.getcwd()
+    for part in directories:
+        path = os.path.join(path, part)
+        if os.path.exists(path):
+            continue
+        os.makedirs(path)
```
viktoria-rybenchuk, line 12: You can achieve the same result without using a for loop by directly calling os.makedirs with the full path you want to create. The `exist_ok=True` parameter ensures that it won't raise an error if the directory already exists.

viktoria-rybenchuk (changes requested):
Screenshot from 2023-11-02 14-00-38

viktoria-rybenchuk:
Screenshot from 2023-11-02 14-01-42
Your code should work the same as in README.md.

app/create_file.py
loplicat: Thanks, I've implemented this changes

viktoria-rybenchuk (approved):
Good job!

PR #583, opened 2023-11-04
Solution

svitlana-savikhina:


app/create_file.py
```diff
@@ -1 +1,30 @@
...
+
+def create_file() -> None:
+    current_time = datetime.datetime.now()
+    if "-d" in sys.argv:
+        for i in range(2, len(sys.argv)):
+            if "-f" not in sys.argv[i]:
+                os.mkdir(sys.argv[i])
+                os.chdir(sys.argv[i])
+            else:
+                break
```
ArsenPidhoretskyi, line 14: Use .index, os.path.join, os.makedirs
Instead of loop, os.mkdir

app/create_file.py
```diff
@@ -1 +1,30 @@
...
+
+    if "-f" in sys.argv:
+        with open(sys.argv[-1], "a") as file:
```
ArsenPidhoretskyi, line 17: But if command will be:

ArsenPidhoretskyi (changes requested):

app/create_file.py
```diff
@@ -1 +1,32 @@
...
+import datetime
+import sys
+import os
+
+
+def create_file() -> None:
+    current_time = datetime.datetime.now()
+    if "-d" in sys.argv:
+        index_d = sys.argv.index("-d")
+        directory_path = sys.argv[index_d + 1:]
```
YukihiroSM, line 10: if the command will be: python create_file.py -d dir1 dir2 dir3 -f file.txt, then this code will use directory path as: 

`dir1 dir2 dir3 -f file.txt` - this will create all there args as separate folders

YukihiroSM: Also, the file is not created inside created folders: 
image

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #584, opened 2023-11-05
Solution

frezworx:


app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        file.write(date_now + "\n")
+        for index, line in enumerate(data):
+            index += 1
+            file.write(f"{index} {line}\n")
+        file.write("\n")
+
+
+def creating_dir(path: str) -> None:
+    if not os.path.exists(path):
+        os.makedirs(path, exist_ok=True)
```
viktoria-rybenchuk, line 32: This code is not redundant, It checks if a directory specified by the path variable exists and, if it doesn't, it creates the directory. The os.makedirs function with `exist_ok=True` ensures that the directory is created if it doesn't exist or does nothing if it already exists.

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+
+
+def get_timestamp() -> str:
+    date_now = datetime.now()
+    return date_now.strftime("%Y-%m-%d %H:%M:%S")
```
viktoria-rybenchuk, line 37: why do you need additional variable?

viktoria-rybenchuk (changes requested):

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+        file.write(date_now + "\n")
+        for index, line in enumerate(data):
+            index += 1
+            file.write(f"{index} {line}\n")
+        file.write("\n")
+
+
+def creating_dir(path: str) -> None:
+    if not os.path.exists(path):
+        os.makedirs(path)
```
Dimosphen1, line 32: Omit this method and use `os.makedirs` where needed.

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+
+    elif "-d" in sys.argv:
+        d_index = sys.argv.index("-d")
+        dir_path = os.path.join(sys.argv[d_index + 1], sys.argv[d_index + 2])
+        creating_dir(dir_path)
+
+    elif "-f" in sys.argv:
+        f_index = sys.argv.index("-f")
+        file_name = sys.argv[f_index + 1]
+        creating_file(file_name)
```
Dimosphen1, line 57: There are several issues with the code:

- You do not call the `create_file` function, it leads to the incorrect script processing. Add `if __name__ == '__main__':` block with this function call;
- You leave two blank lines in a newly created file, there should not be any blank lines at the end of the file after the user finishes the input according to the task description;
- `get_timestamp` function is called only once, you could write the code directly where needed.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

Polyakiv-Andrey (approved):
Good job!

PR #586, opened 2023-11-08
Solution

valikdeb:


checklist.md
```diff
@@ -29,7 +29,7 @@ Good example:
 
 ```python
 def create_path(directories: list) -> str:
-    path = os.path.join(*directories)
+    path = os.path_to_file.join(*directories)
```
Dimosphen1, line 32: Do not make changes in this file, it should be excluded from the PR.

app/create_file.py
```diff
@@ -1 +1,24 @@
...
+
+
+name = argv[f_index + 1]
+with open(path_to_file + "/" + name, "w") as f:
+    f.write(datetime.now().strftime("%Y/%m/%d, %H:%M:%S") + "\n")
+    while True:
+        line = input("Enter content line: ")
+        if line == "stop":
+            break
+        f.write(line + "\n")
```
Dimosphen1, line 24: There are several issues with your code:

- `python create_file.py -d dir1 dir2` command is not working;
- `python create_file.py -f file.txt` command is not working as well;
- Make sure you do not use `/` directory delimiters, you should utilize the `os` to perform all operations regarding the system path so that your code will be compatible across different platforms;
- Utilize functions to distribute the useful functionality and make it reusable;
- Read closely the requirements of the task and example, your script should produce the same results as mentioned there, including the line enumeration, absence of blank lines at the end of a created or updated file, etc.

Dimosphen1 (changes requested):

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+from sys import argv
+from os import makedirs, path
+from datetime import datetime
+
+
+def find_dir() -> str:
+    d_index = argv.index("-d")
+    if "-f" in argv:
+        f_index = argv.index("-f")
+        directory = "/".join(argv[d_index + 1: f_index]) + "/"
```
viktoria-rybenchuk, line 10: Will your code work for another OS?

viktoria-rybenchuk (changes requested):
Your code doesn' work correctly if i call this command 2 times ` python app/create_file.py -f file.txt
`. You should add content, not rewrite it

PR #587, opened 2023-11-09
Solution

IvanMozhar:


app/create_file.py
```diff
@@ -1 +1,62 @@
...
+
+def create_file(file_name: str, path: str) -> None:
+
+    content_file = []
+
+    while True:
+        user_input = input("Enter content line: ")
+        if user_input == "stop":
+            break
+        content_file.append(f"\n{user_input}")
```
viktoria-rybenchuk, line 29: you should add number lines of content as it provided in example

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+
+    if directories:
+        base_directory = os.path.normpath(directories[0])
+
+        os.makedirs(base_directory)
+
+        for directory in directories[1:]:
+            base_directory = os.path.join(base_directory, directory)
+            os.makedirs(base_directory)
+        os.chdir(base_directory)
```
viktoria-rybenchuk, line 18: you can  use `the os.path.join(*directories)` method to create the path by unpacking the list of directories, and then use `os.makedirs` with the exist parameter to create the directories.

app/create_file.py
```diff
@@ -1 +1,62 @@
-# write your code here
+import os
+
+import sys
+
+from datetime import datetime
```
viktoria-rybenchuk, line 5:
```suggestion
import os
import sys
from datetime import datetime
```

viktoria-rybenchuk: the import statements should be organized following PEP 8

viktoria-rybenchuk (changes requested):

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+import sys
+from datetime import datetime
+
+
+def create_directory(directories: list) -> None:
+
+    if directories:
+        base_directory = os.path.join(*directories)
+
+        os.makedirs(base_directory)
```
viktoria-rybenchuk, line 11: Ensure that the base directory is created or exists with `exist=True` parameter

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        os.chdir(base_directory)
```
viktoria-rybenchuk, line 12: why do you need change directory when you have created directories?

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+
+
+def create_file(file_name: str, path: str) -> None:
+
```
viktoria-rybenchuk, line 16: there is no need blank line

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        content_file.append(f"\n{line_number} {user_input}")
+
+    with open(os.path.join(path, file_name), "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+        file.write(" ".join(content_file))
+        file.write("\n\n")
+
+
+if __name__ == "__main__":
+
```
viktoria-rybenchuk, line 34: remove a blank line

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+    command = sys.argv
+
+    if "-d" in command and "-f" not in command:
+
```
viktoria-rybenchuk, line 38: you can remove the blank line for code clarity and to improve the overall structure of the code. Blank lines are typically used to separate different sections of code, but in this case, the blocks are related and follow a specific conditional structure.

viktoria-rybenchuk (changes requested):

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        create_directory(directories)
+    elif "-f" in command and "-d" not in command:
+        f_index = command.index("-f")
+        file_name = command[f_index + 1]
+        path = os.getcwd()
+        create_file(file_name, path)
+    elif "-f" in command and "-d" in command:
+        d_index = command.index("-d")
+        f_index = command.index("-f")
+        directories = command[d_index + 1:f_index]
```
pavlejviki, line 41: Your current code assumes that the directory arguments will come before the file argument, which may not always be the case.

pavlejviki (changes requested):

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+        line_number += 1
+        content_file.append(f"\n{line_number} {user_input}")
+    with open(os.path.join(path, file_name), "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
+        file.write(" ".join(content_file))
+        file.write("\n\n")
+
+
+if __name__ == "__main__":
+    command = sys.argv
```
Polyakiv-Andrey, line 28: Take everything from if main into a separate function and then just call it

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #588, opened 2023-11-09
solution

Andrey1104:


viktoria-rybenchuk (changes requested):
Your doesn't work as expected. For example this command `python create_file.py -d dir1 dir2`  doesn't work. You don't need create content with this command. You just have to create directories

Polyakiv-Andrey (approved):
Good job!

PR #594, opened 2023-11-18
Implemented py-create-file-from-terminal

Aeros72:


app/create_file.py
```diff
@@ -1 +1,65 @@
...
+        file_name = sys.argv.index("-f") + 1
+
+        if dir_index < file_name:
+            directory_path = sys.argv[dir_index:file_name - 1]
+        else:
+            directory_path = sys.argv[dir_index:]
+
+    file_content = ""
+
+    while 1:
```
Dimosphen1, line 54: Use `while True`, it is a common practise to start an infinite loop.

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+        file_name: str,
+        file_content: str
+) -> None:
+
+    if directory:
+        directory_path = os.path.join(*directory)
+        os.makedirs(directory_path, exist_ok=True)
+        path = os.path.join(directory_path, file_name)
+    else:
+        path = file_name
```
Dimosphen1, line 18: You could extract the logic of creating directories to a separate function and call it when needed.

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+        if line == "stop":
+            break
+
+        file_content += line + "\n"
+
+    create_file(directory_path, file_name, file_content)
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 65: There are several issues with your code:

- When you create a new file, there should not be any blank lines after the user finishes the input, currently, you leave one blank line at the end of the file;
- When the user appends new lines to the already created file, there should be one blank line between different sessions of the user input. Currently, there is no blank line between different input sessions;
- When the user runs `python create_file.py -d dir1 dir2 -f file.txt` command, there is an error `TypeError: join() argument must be str, bytes, or os.PathLike object, not 'int'`. Fix that error to create new directories and files correctly in this scenario. 

Besides that, do not use single-letter variable names, be more explicit in naming.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #597, opened 2023-11-22
'Solution'

Illya-Maznitskiy:


app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        index_d = sys.argv.index("-d")
+        index_f = sys.argv.index("-f")
+
+        directory = sys.argv[index_d + 1:index_f]
+        filename = sys.argv[index_f + 1]
+
+        # Call the function with descriptive names
+        create_file(directory, filename)
+    else:
+        print("Usage: python create_file.py -d <directory> -f <filename>")
```
Dimosphen1, line 58: There are several issues with your code:

- Your command should run when only the `-d` flag is passed;
- Your command should run when only the `-f` flag is passed;
- You should enumerate lines with user input according to the task description;
- You should have a file structure as mentioned in the task description in case user performs several interactions of input over one file.

Besides that, remove comments from your code, there is no need for them.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #599, opened 2023-11-24
Solution

VoloIvanchuk:


app/create_file.py
```diff
@@ -1 +1,34 @@
...
+        while text != "stop":
+            new_file.write(text + "\n")
+            text = input("Enter content line: ")
+        new_file.write("\n")
+
+
+if __name__ == "__main__":
+    command = sys.argv
+
+    try:
```
vasylhnatiuk, line 27: all of logic must be before `if __name__ == "__main__"`
here you must only call function

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #639, opened 2024-01-21
Solution

ArturPoltser:


aLEKS-e3 (approved):

NEkropo1 (approved):
You could create subclasses of raisen Error, or just use ValueError, anyway, good job!

PR #640, opened 2024-01-21
Solution

aLEKS-e3:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+def get_clean_command(cmd: str) -> str:
+    return " ".join(sys.argv).split(cmd)[1]
+
+
+def file_creation(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        line = 1
+        file.write(
+            datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n"
+        )
```
NEkropo1, line 15:
```suggestion
        current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
        file.write(f"{current_time}\n")
```

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+        if new_file:
+            file_path = os.path.join(path, new_file)
+    elif "-f" in sys.argv:
+        file_path = get_clean_command("-f")
+    else:
+        raise ValueError("No '-f' or '-d' command was passed!")
+
+    if file_path:
+        file_creation(file_path)
```
NEkropo1, line 44: else is excessive, you could handle it this way, or similar
```suggestion
    if not file path:
        raise ValueError("No '-f' or '-d' command was passed!")

    file_creation(file_path)
```

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+import os
+import sys
+from datetime import datetime
+
+
+def get_clean_command(cmd: str) -> str:
+    return " ".join(sys.argv).split(cmd)[1]
+
+
+def file_creation(file_path: str) -> None:
```
NEkropo1, line 10:
```suggestion
def create_file(file_path: str) -> None:
```
Suggeesting this name, because it seems more appropriate, when you use it in other functions, like function created file, not simply calls for file creation... Just a suggestion here)

NEkropo1 (changes requested):

app/create_file.py
aLEKS-e3: ok, but won't it create confusion because of the same filename?

app/create_file.py
aLEKS-e3: I don't think so, because if-elif-else statement here checks what exactly to create - a directory path or a file one. So basically that `raise` statement tells when no command was passed. 
If I move it to the `if file_path` block then the user will get an error when he/she creates just a directory path.

Hope you got my idea)

PR #642, opened 2024-01-22
Solution

msymonovych:


kabust (approved):

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+            if key_d < key_f:
+                dirs = parameters[key_d + 1: key_f]
+            else:
+                dirs = parameters[key_d + 1:]
+            path = os.path.join(create_dirs(dirs), parameters[key_f + 1])
+            create_file(path)
+
+        else:
+            dirs = parameters[key_d + 1:]
+            create_dirs(dirs)
```
NEkropo1, line 44: try to break it into more readable blocks, without so much nestng, like `if key_d and key_f ifel key_d and not key_d` etc

NEkropo1 (changes requested):

PR #644, opened 2024-01-24
Solution

u123dev:


app/create_file.py
```diff
@@ -1 +1,83 @@
...
+    py create_file.py -d dir1 [ .. dirN] [-f filename.ext]
+        or
+    py create_file.py -f filename.ext [-d dir1 [ .. dirN]]
+
+Enter line by line to file from the terminal. To complete - enter "stop" """)
+
+
+def parse_cmd(list_args: list) -> None:
+    # output:  fpath = {"dirs": "", "filename": ""}
+    fpath = {}
```
NikOneZ1, line 19: It is not the best practice to name variables like this, better use name like file_path

app/create_file.py
```diff
@@ -1 +1,83 @@
...
+
+    idx_f = len(list_args)
```
NikOneZ1, line 21: Same here, it would be better to name this variable like file_name_index

NikOneZ1 (changes requested):

NikOneZ1 (approved):

PR #645, opened 2024-01-24
Solution

Acik33:


vsmutok (changes requested):
running command a second time raise an error

image

PR #647, opened 2024-01-24
Solution

Vladyslav-Pliashechnyk:


aLEKS-e3 (approved):

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    command = argv
+    if ("-d" in command) and ("-f" not in command):
+        create_directory(command[2:])
+    elif ("-f" in command) and ("-d" not in command):
+        create_file(command[2])
+    elif ("-d" in command) and ("-f" in command):
+        arguments = [arg for arg in command[2:] if arg != "-f"]
+        create_directory(arguments[:-1])
+        create_file(file_path(arguments))
+    else:
```
NEkropo1, line 38: You need this else?)

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+        line_number = 1
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{timestamp}\n")
+        while True:
+            content = input("Enter content line: ")
+            if content == "stop":
+                file.write("\n")
+                break
+            file.write(f"{line_number} {content}\n")
+            line_number += 1
```
NEkropo1, line 25:
```suggestion

        while True:
            content = input("Enter content line: ")
            if content == "stop":
                file.write("\n")
                break
            file.write(f"{line_number} {content}\n")
            line_number += 1
```

app/create_file.py
```diff
@@ -1 +1,43 @@
-# write your code here
+from sys import argv
+from os import path, makedirs
+from datetime import datetime
```
NEkropo1, line 3: a simple alphabetical order, please

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+
+
+def file_path(arguments: list) -> str:
+    return path.join(*arguments)
```
NEkropo1, line 7: common way will be importing whole os module and use os.path.join explicitly, since we can have multiple variables like path and join, so we can make simple mistakes.

NEkropo1 (changes requested):

PR #650, opened 2024-01-26
Solved

Dmitriy-Poplinski:


vsmutok (changes requested):
`python create_file.py -d dir1 dir2` not work
image

vsmutok (approved):
Great!

PR #651, opened 2024-01-26
Update create_file.py

opminen:


app/create_file.py
```diff
@@ -1 +1,31 @@
...
+commands_list = sys.argv[1:]
+
+dirs = []
+user_file = commands_list[commands_list.index("-f") + 1:][0]
+
+if "-d" in commands_list:
+    for el in commands_list[1:]:
+        if el == "-f":
+            break
+        dirs.append(el)
```
YukihiroSM, line 14: What if -f option is before -d option? Then you will break dirs collection at the very beginning of the program and you will just create a file without directories. 
Can you think about this as well?

app/create_file.py
```diff
@@ -1 +1,31 @@
...
+
+path_for_file = os.path.join(*dirs)
+
+if os.path.isdir(path_for_file):
+    is_file_exist = True
+else:
+    is_file_exist = False
+    os.makedirs(path_for_file)
```
YukihiroSM, line 22: You may get exception if directories already exist. Add exists_ok parameter, please.

YukihiroSM (changes requested):

PR #653, opened 2024-01-28
Solution

Ded31:


app/create_file.py
```diff
@@ -1 +1,26 @@
...
+        directories = command[2:command.index("-f")]
+    except ValueError:
+        directories = command[2:]
+
+    dir_path = os.path.join(*directories)
+    os.makedirs(dir_path, exist_ok=True)
+
+
+if "-f" in command:
+    with open(os.path.join(dir_path, command[-1]), "w") as f:
```
YukihiroSM, line 21: It will not work for the case, when the command is like: 
python copy_file.py -f file.txt -d dir1/dir2

YukihiroSM (changes requested):
Also, I'm getting an exception trying to execute your code with the following parameters:
python create_file.py -f test.txt -d test/dir

Other thing: there is no lines numbering. 

According to the task you should number for each line. 

image

vasylhnatiuk (approved):

PR #655, opened 2024-01-29
Solution

Vityk-4W:


app/create_file.py
```diff
@@ -1 +1,38 @@
-# write your code here
+import os
+import sys
+import datetime
+
+
+if __name__ == "__main__":
```
vasylhnatiuk, line 6: after this line better write only function calling, not all the code

vasylhnatiuk (changes requested):
separate your code to a few functions with different responsibilities

app/create_file.py
vasylhnatiuk, line 6: after this line better write only function calling, not all the code

vasylhnatiuk (changes requested):

EdAlekseiev (approved):
Godo job

PR #656, opened 2024-01-29
Solution

KwintDimon:


app/create_file.py
```diff
@@ -1 +1,53 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_folder(folders: list[str]) -> str:
+    path = os.path.join(*folders)
+    if not os.path.exists(path):
+        os.makedirs(path)
```
Dimosphen1, line 9: You could use the `exist_ok` flag in the `makedirs` instead of the if condition.

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+            create_folder(dirs)
+
+        else:
+            dirs = (
+                args[flag_d + 1: flag_f]
+                if flag_f > flag_d
+                else args[flag_d + 1:]
+            )
+            path = os.path.join(create_folder(dirs), args[flag_f + 1])
+            create_file(path)
```
Dimosphen1, line 49: You could make a single-level condition without nesting. Rewrite this part without using nesting in the `elif flag_d:`.

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 53: There are several issues with your code:
- The line enumeration is not correct, there is only line 1 present (no incrementation); 
- You should place only the line number, and after that, the input content, word `Line` and the following number are redundant (refer to the example);
- You should not leave two blank lines at the end of the file after the end of the input session, there should be no blank lines at the end according to the task description.

Dimosphen1 (changes requested):

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+        file_name = args[flag_f + 1]
+        create_file(file_name)
+
+    else:
+        dirs = args[flag_d + 1:]
+        create_folder(dirs)
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 51: There are several issues with your code:
- There is a redundant blank line at the end of the file after exiting the script, there should be no blank lines left at the end of the file;
- In case of appending to a created file, there is no blank line between different input sessions.

Dimosphen1 (changes requested):

PR #661, opened 2024-01-31
Develop

MaxOperchuk:


app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        file.write(f"{formatted_datetime}\n")
+        number_of_line = 1
+        while text != "stop":
+
+            if text != "stop":
+                file.write(f"{number_of_line} {text}\n")
+            number_of_line += 1
+            text = input("Enter content line: ")
+
+        file.write("\n")
```
VladKyianytsia, line 41: Maybe it can be refactored a little bit

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+
+
+if "-d" in command and "-f" in command:
+    way_to_file = create_path(command[2:-2])
+    make_dir(way_to_file)
+
+    path_to_create_file = way_to_file + "/" + command[-1]
```
VladKyianytsia, line 48: You can use os.path.join here

VladKyianytsia (approved):

PR #668, opened 2024-01-31
solution

PythonZem:


tskopenko (approved):
LGTM!~

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+            count = 1
+
+            while True:
+                user_input = input("content line: ")
+                if user_input != "stop":
+                    file_in.writelines(f"{count} {user_input}\n")
+                    count += 1
+                else:
+                    file_in.writelines("\n")
+                    break
```
NEkropo1, line 38: invert this logic to remove else.

app/create_file.py
```diff
@@ -1 +1,47 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+def get_directory_list(commands: list) -> list[str] | None:
+    if commands.index("-d") < len(commands) - 1:
+        return commands[commands.index("-d") + 1: commands.index("-f") | 0]
```
NEkropo1, line 8: if we won't have -f here, it will raise a Value error.

NEkropo1 (changes requested):

PR #669, opened 2024-01-31
solution

motivated2:


AntonSkyrda (changes requested):
command python create_file.py -f file.txt does not create a file.txt

PR #673, opened 2024-01-31
Solution

Nick-r2c:


VladKyianytsia (approved):
Try to test if -f will be before -d flag

NEkropo1 (changes requested):
Does it work as we want?
App should add current timestamp at the top and number lines. If file.txt already exists it should add content below:

python create_file.py -f file.txt

Nick-r2c:
> Does it work as we want? App should add current timestamp at the top and number lines. If file.txt already exists it should add content below:
> 
> python create_file.py -f file.txt

I believe, now it works flawlessly

PR #678, opened 2024-02-02
Create file from terminal

Ananiev-Vitalii:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+def main() -> None:
+    args = sys.argv[1:]
+
+    if "-d" in args:
+        dir_index = args.index("-d")
+        directory = os.path.join(*args[dir_index + 1:])
+        os.makedirs(directory, exist_ok=True)
+    else:
+        directory = ""
```
EdAlekseiev, line 30: This doesn't work for `python create_file.py -d dir1 dir2 -f file.txt`

EdAlekseiev (changes requested):

PR #695, opened 2024-02-11
Solution

kapitoshk4:


app/create_file.py
```diff
@@ -1 +1,51 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory_if_not_exists(directory_path: str) -> None:
+    if not os.path.exists(directory_path):
+        os.makedirs(directory_path)
```
NikOneZ1, line 8: You can use os.makedirs exist_ok=True

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #698, opened 2024-02-15
Solution

mykytaso:


vasylhnatiuk (changes requested):
Знімок екрана 2024-02-15 о 14 52 58  
it doesn't create folders with command `python create_file.py -d dir1 dir2 -f file.txt`

vasylhnatiuk (changes requested):
try to separate your code to a few functions and after it call main functions

PR #704, opened 2024-02-19
Solution

vsupruniuk:


app/create_file.py
```diff
@@ -1 +1,59 @@
-# write your code here
+from datetime import datetime
+import os
+import sys
+
+
+def create_dirs(path: sys) -> None:
+    if not os.path.exists(path):
+        os.makedirs(path)
+
```
LiudmylaKulinchenko, line 9: u can use makedirs argument "exist_ok" here

LiudmylaKulinchenko: then there won't be need to create a new function to use another one

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+
+def create_file_from_input(dir_path: str, file_name: str) -> None:
+    with open(os.path.join(dir_path, file_name), "w") as file:
+        content_lines = [datetime.now().strftime("%Y-%m-%d %H:%M:%S")]
+
+        while True:
+            line = input("Enter content line:")
+
+            if line.lower() == "stop":
```
LiudmylaKulinchenko, line 18: you can make this check directly in `while`

LiudmylaKulinchenko (changes requested):
you can also use argparse module here — it would be easier, more readable, and you won't have to "reinvent the wheel"

LiudmylaKulinchenko (approved):
Nice job! 🍬🌟

PR #706, opened 2024-02-20
solution: create file from terminal

sacrosanctry:


app/create_file.py
```diff
@@ -1 +1,51 @@
...
+
+if __name__ == "__main__":
+    arguments = sys.argv
+    path_list, file_name = parse_arguments(arguments)
+
+    if path_list:
+        create_dirs(path_list)
+
+    if file_name:
+        create_file(path_list, file_name)
```
Dimosphen1, line 51: Do not add blank lines at the end of a newly created file, the content should be the same as one from the task description

Dimosphen1 (changes requested):

PR #709, opened 2024-02-21
Provided the implementation of the assigned task

raychw:


app/create_file.py
```diff
@@ -1 +1,57 @@
...
+    file_name = None
+
+    if "-d" in sys.argv:
+        dir_index = sys.argv.index("-d") + 1
+        dir_args = []
+
+        for arg in sys.argv[dir_index:]:
+            if arg.startswith("-f"):
+                break
+            dir_args.append(arg)
```
vsmutok, line 17: try to use  list slicing instead of a loop

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+
+    with open(file_path, "a") as f:
+
+        if file_exists:
+            f.write("\n")
+
+        f.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
+
+        if file_exists:
+            line_num = sum(1 for _ in open(file_path)) + 1
```
vsmutok, line 45: this is redundant here 
if there is already data in the file, the numbers should all start from the beginning
image

vsmutok (changes requested):
Also use `balck` to fix indentations

Polyakiv-Andrey (approved):
Good job!

PR #715, opened 2024-02-28
Solution

VasylRybalka:


app/create_file.py
```diff
@@ -1 +1,36 @@
...
+        file.write("\n")
+
+
+def finding_path() -> None:
+    cwd = sys.argv
+    to_dir = ""
+
+    if cwd[1] == "-d":
+        if "-f" not in cwd:
+            to_dir = os.path.join(*cwd[2:])
```
EdAlekseiev, line 26: It doesn't work for `python create_file.py -d dir1 dir2 -f file.txt`

EdAlekseiev (changes requested):

PR #727, opened 2024-03-08
Solution

vburkalo:


app/create_file.py
```diff
@@ -1 +1,65 @@
...
+        file_name: str,
+        content_lines: List[str]
+) -> None:
+    if directory_path:
+        os.makedirs(directory_path, exist_ok=True)
+
+    if file_name:
+        full_path = (
+            os.path.join(directory_path, file_name)) \
+            if directory_path else file_name
```
YukihiroSM, line 38: Don't use `\` for line separation. It's better to use `(` `)`

YukihiroSM (changes requested):
That's the only tiny comment about line separation. Else looks good. 

Also you are not reuqired to create tests here.

EdAlekseiev (approved):
Perfect job

PR #732, opened 2024-03-11
Develop

vanko6226:


app/create_file.py
```diff
@@ -1 +1,53 @@
...
+            while input_str != "stop":
+                counter += 1
+                input_str = input("Enter content line: ")
+                if input_str != "stop":
+                    f.write(f"{counter} {input_str}\n")
+            f.write("\n")
+
+
+if __name__ == "__main__":
+    command_line_args = sys.argv[1:]
```
vasylhnatiuk, line 50: what if command is worng and sys.argv don't have a few arguments? you should inform user, if command doesn't have all arguments

vasylhnatiuk (changes requested):

NikOneZ1 (approved):

PR #736, opened 2024-03-19
Solution

halyna-baklanova:


pavlopro (changes requested):
If the file already exists, it should add content below the existing content. Currently, it overwrites existing content.

vasylhnatiuk (approved):

PR #740, opened 2024-03-21
Solution

nicksetrakov:


app/create_file.py
```diff
@@ -1 +1,53 @@
...
+
+def create_dir_or_file(
+        directory: list[str] | None, file_name: list[str] | None
+) -> None:
+
+    if directory:
+        directory_path = os.path.join(*directory)
+        os.makedirs(directory_path, exist_ok=True)
+    else:
+        directory_path = ""
```
Oleksl888, line 15: It would be better to use `os.getcwd()` method to get absolute path to current directory

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+        write_to_file(file_path)
+
+
+def write_to_file(file_path: str) -> None:
+    count = 1
+
+    with open(file_path, "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+
+        while True:
```
Oleksl888, line 28: opening an infinite loop inside open file handle is not optimal. 

 It is better to open a loop, collect all data in a separate function.

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+            content = input("Enter content line: ")
+
+            if content.lower() == "stop":
```
Oleksl888, line 31: If you follow the previous step and separate gathering input from user think what will be the most optimal way to prepare data before passing it to your `write_to_file` function.

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+                file.write("\n")
+                break
+
+            file.write(str(count) + " " + content + "\n")
```
Oleksl888, line 35: How can you substitute writing str(count) and have alternative to write data in a loop  on each iteration

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+            count += 1
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
```
Oleksl888, line 40: Why did you choose this method other then `argv`?

app/create_file.py
```diff
@@ -1 +1,53 @@
...
+) -> None:
+
+    if directory:
+        directory_path = os.path.join(*directory)
+        os.makedirs(directory_path, exist_ok=True)
+    else:
+        directory_path = ""
+
+    if file_name:
+        file_path = os.path.join(directory_path, *file_name)
```
Oleksl888, line 18: what if directory's path/name will be change or deleted right before this step?

Oleksl888 (changes requested):

app/create_file.py
Oleksl888: > what if directory's path/name will be change or deleted right before this step?

sorry I meant line 25

YuriiKindrat (approved):
LGTM

PR #744, opened 2024-03-22
Implement solution

Lumideia:


app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        directory_path, f_flag_position = read_file_path()
+    else:
+        file_name = read_file_name()
+    if f_flag_position != argv_length - 1:
+        file_name = read_file_name()
+
+    path = ""
+    if directory_path:
+        path = os.path.join(*directory_path)
+        os.makedirs(path)
```
EdAlekseiev, line 56: Use exist_ok=True to avoid errors
```suggestion
        os.makedirs(path, exist_ok=True)
```

EdAlekseiev (changes requested):

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+    return file_path, next_flag_position
+
+
+def read_file_name() -> str:
+    return sys.argv[argv_length - 1]
+
+
+def read_input() -> list:
+    content = []
+    while 1:
```
vasylhnatiuk, line 24:
```suggestion
    while True:
```

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        new_line = input("Enter content line: ")
+        if new_line == "stop":
+            break
+        content.append(new_line)
+    return content
+
+
+def write_info(file_to_write_info: str, content: list) -> None:
+    with open(file_to_write_info, "a") as f:
```
vasylhnatiuk, line 33: f is a bad name

app/create_file.py
```diff
@@ -1 +1,58 @@
...
+        f.write(f"{datetime.datetime.now().strftime('%I:%M%p %d/%B/%Y')}\n")
+        for i in range(len(content)):
```
vasylhnatiuk, line 35: use more accurate name instead of `i`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #745, opened 2024-03-23
Solution

Bujhm99:


nicksetrakov (changes requested):
Your code works correctly only when it receives both flags -d and -f

photo_2024-03-23_16-50-31

photo_2024-03-23_16-50-55

photo_2024-03-23_16-52-04

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+try:
+    flag_dir_pos = current_messege.index("-d")
+except ValueError:
+    pass
+else:
+    while (flag_dir_pos < len(current_messege) - 1
+           and current_messege[flag_dir_pos + 1] != "-f"):
+        new_path = os.path.join(new_path, current_messege[flag_dir_pos + 1])
+        if not os.path.exists(new_path):
+            os.mkdir(new_path)
```
YuriiKindrat, line 44: try to use mkdirs instead with arg (exists_ok)

YuriiKindrat (changes requested):
make ur code more modular, it is really hard to read it

app/create_file.py
```diff
@@ -1 +1,50 @@
...
+                                        next_messege)
+                os.makedirs(new_path, exist_ok=True)
+                flag_position += 1
+
+
+try:
+    _ = open(os.path.join(new_path, new_file_name), "x")
+except IOError:
+    pass
+write_info(os.path.join(new_path, new_file_name))
```
YuriiKindrat, line 50: it is better to create separate methods for every thing u do )
as for me, u could create 2 more funct here, for parse args, and create dirs 
also u could create main func to call other funcs inside it 
Also try to use argparse for parsing arguments from terminal - this will simplify ur code, and make it easier to read )

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+            counts += 1
+
+
+new_path = os.path.dirname(os.path.abspath(__file__))
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-f", nargs=1, dest="file_work", required=True,
+                    help="Only 1 name of file to write")
+parser.add_argument("-d", nargs="*", dest="dicts", default=None)
+parser.add_argument("baz", nargs="*")
```
Oleksl888, line 31: what this is doing?

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+            line = input("Enter content line: ")
+            counts += 1
+
+
+new_path = os.path.dirname(os.path.abspath(__file__))
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-f", nargs=1, dest="file_work", required=True,
+                    help="Only 1 name of file to write")
+parser.add_argument("-d", nargs="*", dest="dicts", default=None)
```
Oleksl888, line 30: `python create_file.py -d dir1 dir2 - creates directory dir1/dir2 inside current directory.` Your code doesn't do that

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+parser.add_argument("baz", nargs="*")
+argss = parser.parse_args(sys.argv)
+
+if argss.dicts:
+    new_path = os.path.join(new_path, os.path.sep.join(argss.dicts))
+os.makedirs(new_path, exist_ok=True)
```
Oleksl888, line 36: your code doesn't handle wrong input (when dir/file names contain slashes or other unacceptable symbols)

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+
+def write_info(file_name: str) -> None:
+
+    line = input("Enter content line: ")
+    with open(file_name, "a") as file_open:
+        if os.stat(file_name).st_size:
+            file_open.write("\n")
+
+        counts = 1
+        while line != "stop":
```
Oleksl888, line 15: Creating infinite loop inside open file is not an optimal solution in a real world scenario. It is better to take user input in a separate function and then pass the result to your write_info method.

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+            if counts == 1:
+                file_open.write(datetime.datetime.now()
```
Oleksl888, line 17: would be more readable to break lines like this 
```
file_open.write(
    datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n"
)

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: app/create_file.py (hunk 3 above)
Comment (Bujhm99) [reply app/create_file.py:31]: "baz" is for positional arguments

File: app/create_file.py (hunk 4 above)
Comment (Bujhm99) [reply app/create_file.py:30]: According condition create_file.py placed in .app, so i use "python app\create_file.py" and all creating in it.

**Review by YuriiKindrat — CHANGES_REQUESTED** [review]
GJ, but create a separate func for arg parcing and and calling write_info method 
e.g. 'main' 
and call it with name==main

File: app/create_file.py [hunk 8]
```diff
@@ -1 +1,59 @@
...
+    if argss.dicts:
+        new_path = os.path.join(new_path, os.path.sep.join(argss.dicts))
+
+        for char in '|*?"</>':
+            if char in new_path:
+                raise OSError(f"Wrong char {char} in command dir")
+
+    for char in ':|*?"</>':
+        if char in argss.file_work[0]:
+            raise OSError(f"Wrong char {char} in file")
```
Oleksl888, line 47: Would be more optimal to wrap create dirs / files into try-except and catch OSError there - right now the complexity of your code is n*2 - first you iterate in a for loop and inside for loop you perform naive search algorythm which is results in quadratic complexity

app/create_file.py
```diff
@@ -1 +1,59 @@
-# write your code here
+import argparse
+import sys
+import os
+import datetime
+
+
+def write_info(file_name: str) -> None:
+
+    line = input("Enter content line: ")
```
Oleksl888, line 9: you could structure this differently to avoid repeating this line

Oleksl888 (approved):
Approved

PR #747, opened 2024-03-26
Develop

saywin:


app/create_file.py
```diff
@@ -1 +1,32 @@
...
+        number_of_lines = 1
+
+        while file_contents != "stop":
+            f.write(f"{number_of_lines} {file_contents}\n")
+            file_contents = input("Enter content line: ")
+            number_of_lines += 1
+        f.write("\n")
+
+
+def create_data_for_file() -> None:
```
vasylhnatiuk, line 20:
```suggestion
def make_dirs_and_create_file() -> None:
```

vasylhnatiuk: it is more accurate name, but would be better if you create separate function for creating dirs

vasylhnatiuk (changes requested):

app/create_file.py
saywin: Done

vasylhnatiuk (approved):

saywin:
Thank you

PR #753, opened 2024-04-02
Solution

Line181:


app/create_file.py
```diff
@@ -1 +1,52 @@
...
+    with open(new_path, "a") as new_file:
+        new_file.write(text)
+
+
+def main() -> None:
+    parse_terminal()
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 52: - Your lines should be enumerated in the final file as mentioned in the task description;
- Date formatting should be the same as mentioned in the task (using `:` instead of `-` for time).

Dimosphen1 (changes requested):

PR #761, opened 2024-04-04
Solution

vitalinamal:


app/create_file.py
```diff
@@ -1 +1,77 @@
...
+
+
+def create_directory(path_parts: list[str]) -> str:
+    try:
+        directory_path = os.path.join(*path_parts)
+        os.makedirs(directory_path, exist_ok=True)
+        return directory_path
+    except PermissionError:
+        print(f"Permission denied: "
+              f"unable to create directory '{directory_path}'.")
```
Oleksl888, line 23: directory_path is referenced before assignment - line 18 may fail and directory_path variable will not be created but you try to call it in exception message

app/create_file.py
```diff
@@ -1 +1,77 @@
...
+    except Exception as e:
+        print(f"An error occurred creating or "
+              f"writing to file '{file_path}': {e}")
+
+
+def main() -> None:
+    try:
+        args = sys.argv[1:]
+
+        if "-d" in args and "-f" in args:
```
Oleksl888, line 57: what if params come in the wrong order?

Oleksl888 (changes requested):

YuriiKindrat (changes requested):
It should work fine, but try to use Argparcer to simplify ur solution with args parcing

LiudmylaKulinchenko (approved):
Good job 🌷

PR #764, opened 2024-04-05
Solution

panicua:
An interesting task. I used ```argparse``` instead of ```sys.argv``` cause in video the lecturer said it's used more often for work with terminal so I googled some about this lib and decided to use it. Its method ```ArgumentParser``` has lots of arguments that you can change in order to interact with the commands you wrote in terminal. 👍

### DIFF
--- a/app/create_file.py
+++ b/app/create_file.py
@@ -1,1 +1,70 @@
-# write your code here
+import argparse
+import os
+import datetime
+
+
+def terminal_to_file() -> None:
+    parser = argparse.ArgumentParser(
+        description="The app creates folders/files, "
+                    "writes inside of those files"
+    )
+    parser.add_argument("-d",
+                        nargs="+",
+                        default=[],
+                        help="'-d dir1 dir2' - creates directory dir1/dir2")
+    parser.add_argument("-f",
+                        type=str,
+                        default=None,
+                        help="'-f file.txt' - creates 'file.txt' "
+                             "in current directory")
+
+    try:
+        args = parser.parse_args()
+    except argparse.ArgumentError:
+        raise argparse.ArgumentError
+
+    d_arguments, f_argument = args.d, args.f
+
+    if d_arguments:
+        create_folder(d_arguments)
+    if f_argument:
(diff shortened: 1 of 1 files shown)


panicua:
had to add ```Q003``` to ```flake8``` ignore list cause it didn't let me use beautiful output inside double quotes so I don't need to use single once: 
these: ```help="\"-d dir1 dir2\" - creates directory \"dir1/dir2\"."``` < BEAUTIFUL DOUBLE QUOTES
instead of this: ```help="'-d dir1 dir2' - creates directory 'dir1/dir2'."``` < UGLY SINGLE QUOTES

panicua:
returned flake8 ```.ignore``` file to the starting condition

panicua:
Switched from ```file.write(str)```, to ```file.writelines(list)``` cause when I asked GPT about how can I improve my code, it said that creating a new string everytime: ```full_message += message``` isn't as efficient as to just ```full_message.append(message)```

app/create_file.py
```diff
@@ -1 +1,78 @@
...
+        message = input("content line: ")
+        if message == "stop":
+            full_message.append("\n")
+            break
+        full_message.append(message + "\n")
+
+    if len(full_message) < 2:
+        return print("You cannot add an empty message")
+
+    with open(path, mode) as file:
```
Oleksl888, line 71: you should add error handling here too e.g. in case of file name being wrong

app/create_file.py
```diff
@@ -1 +1,78 @@
...
+
+def writing_to_file(path: str, mode: str) -> None:
+    full_message = list()
+
+    while True:
+        message = input("content line: ")
+        if message == "stop":
+            full_message.append("\n")
+            break
+        full_message.append(message + "\n")
```
Oleksl888, line 66: this can be moved to a separate function as per single responsibility pricinciple

Oleksl888 (changes requested):

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🔥

PR #767, opened 2024-04-07
Solution

SerhiiMar:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+        while True:
+            line_count += 1
+            line_input = input("Enter content line: ")
+            if line_input == "stop":
+                break
+            file.write(f"{line_count} {line_input}\n")
+
+
+arguments = sys.argv
+create_file(arguments, create_folder(arguments))
```
YuriiKindrat, line 43: use name==main here
also it is better to create separate func e.g. 'main' to call this methods

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #769, opened 2024-04-08
Solution

Gilgumish:


YuriiKindrat (changes requested):
make ur code more modular, it is really hard to read it now
split ur logic on the separate functions

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+def main() -> None:
+    args = sys.argv[1:]
+    dir_path, file_name = parse_arguments(args)
+
+    if not file_name:
+        file_name = "file.txt"
+
+    full_dir_path = create_directory(dir_path)
+    full_path = os.path.join(full_dir_path, file_name) \
+        if full_dir_path else file_name
```
LiudmylaKulinchenko, line 55: Avoid using `\` as a linebreaker

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+    content_lines = collect_content_lines()
```
LiudmylaKulinchenko, line 56: U can skip creating this variable

LiudmylaKulinchenko (changes requested):
try to use Argparcer to simplify ur solution with args parcing

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+        for i, line in enumerate(content_lines, start=1):
+            f.write(f"{i}. {line}\n")
+
+
+def main() -> None:
+    args = sys.argv[1:]
+    dir_path, file_name = parse_arguments(args)
+
+    if not file_name:
+        file_name = "file.txt"
```
Oleksl888, line 51: You should not set a default file name in case of its missing. If the input is incorrect raise a warning or print error message and exit program

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+def create_directory(dir_path: str) -> str:
+    if dir_path:
+        full_dir_path = os.path.join(*dir_path)
+        os.makedirs(full_dir_path, exist_ok=True)
+        return full_dir_path
+    return ""
+
+
+def collect_content_lines() -> list:
+    print("Enter content lines (type 'stop' to finish):")
```
Oleksl888, line 27: Must be `Enter content line:` nothing else

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+        content_lines.append(line)
+        line = input()
+    return content_lines
+
+
+def write_content_to_file(full_path: str, content_lines: list) -> None:
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    with open(full_path, "a") as f:
+        if os.path.getsize(full_path) > 0:
+            f.write("\n")
```
Oleksl888, line 40: you can avoid using many writes here if you prepare your data in a list in collect_content_lines function and then use writelines() method once. And it will actually be faster

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+    line = input()
+    while line != "stop":
+        content_lines.append(line)
+        line = input()
+    return content_lines
+
+
+def write_content_to_file(full_path: str, content_lines: list) -> None:
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    with open(full_path, "a") as f:
```
Oleksl888, line 38: you should also handle errors that may occur during working with file

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+        line = input()
+        if line == "stop":
+            break
+        content_lines.append(line)
+    return content_lines
+
+
+def write_content_to_file(full_path: str, content_lines: list) -> None:
+    try:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
Oleksl888, line 39: You could add timestamp and enumerate in your collect_content_lines functions + add all necessary EOL symbols so in this function you could just pass prepared data to writelines().

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+
+def main() -> None:
+    args = sys.argv[1:]
+    dir_path, file_name = parse_arguments(args)
+
+    if not file_name:
+        print("Error: No file name provided.")
+        sys.exit(1)
+
+    full_dir_path = create_directory(dir_path)
```
Oleksl888, line 58: `python create_file.py -d dir1 dir2 - creates directory dir1/dir2 inside current directory.` you code doesn't do it - it throws an error

app/create_file.py
```diff
@@ -1 +1,66 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def parse_arguments(args: list) -> tuple:
+    try:
+        dir_path = args[args.index("-d") + 1:args.index("-f")]
+    except ValueError:
```
Oleksl888, line 9: you should add error messages to your exceptions

Oleksl888 (approved):
Approved. But you need to fix your code. please check notes

PR #771, opened 2024-04-08
Solution

DmytroHlazyrin:


app/create_file.py
```diff
@@ -1 +1,62 @@
...
+            os.makedirs(dir_path, exist_ok=True)
+
+        file_path = os.path.join(dir_path, sys.argv[f_index + 1])
+
+        data_input(file_path)
+
+
+with_d_and_f()
+only_d()
+only_f()
```
YuriiKindrat, line 62: use if name == main here
also it is better to create function e.g. 'main' to call all this functions

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+        list_d = []
+
+        list_d.extend(sys.argv[d_index + 1:f_index])
+        dir_path = os.path.join(*list_d)
+        if dir_path:
+            os.makedirs(dir_path, exist_ok=True)
+
+        file_path = os.path.join(dir_path, sys.argv[f_index + 1])
+
+        data_input(file_path)
```
YuriiKindrat, line 57: try to use Argparces for this, it will much simplify ur code

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+            with open(file_path, "a") as file:
+                file.write(f"{count} {text} \n")
+
+
+parser = argparse.ArgumentParser(description="Create file with input")
+
+parser.add_argument("-d", nargs="+", dest="dir_path")
+parser.add_argument("-f", dest="file_name")
+
+args = parser.parse_args()
```
YuriiKindrat, line 44: could u place this inside main function ?

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+            data_input(file_path)
+        else:
+            data_input(file_name)
+
+
+def data_input(file_path: str) -> None:
+    count = 0
+    if os.path.exists(file_path):
+        with open(file_path) as file:
+            count = len(file.readlines()) - 1
```
YuriiKindrat, line 22: u could move this to the separate method 'get_line_number'

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+        text = input("Enter content line: ")
+        if text.lower() == "stop":
+            break
+        if not os.path.exists(file_path):
+            with open(file_path, "w") as file_time:
+                file_time.write(utc_time + "\n")
+                file_time.write(f"{count} {text} \n")
+        else:
+            with open(file_path, "a") as file:
+                file.write(f"{count} {text} \n")
```
YuriiKindrat, line 36: whey do u need to open file every iteration inside while loop, could u optimize this ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+
+    parser.add_argument("-d", nargs="+", dest="dir_path")
+    parser.add_argument("-f", dest="file_name")
+
+    args = parser.parse_args()
+    dir_path = args.dir_path
+    file_name = args.file_name
+
+    if dir_path:
+        path = os.path.join(*dir_path)
```
Oleksl888, line 17: you can put put creating dir and creating file in separate functions to follow single responsibility pricniple

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+
+
+def data_input(file_path: str) -> None:
+    count = 0
+    utc_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    mode = "a" if os.path.exists(file_path) else "w"
+    with open(file_path, mode) as file_time:
+        file_time.write(f"{utc_time}\n")
+
+        while True:
```
Oleksl888, line 34: would be more optimal to take data from user in separate function before you open the loop to write to file, collect all data and then pass it to write to file

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+            file_path = os.path.join(*dir_path, file_name)
+            data_input(file_path)
+        else:
+            data_input(file_name)
+
+
+def data_input(file_path: str) -> None:
+    count = 0
+    utc_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    mode = "a" if os.path.exists(file_path) else "w"
```
Oleksl888, line 30: you can always use "a" mode - it will also create a file if it doesnt exist

app/create_file.py
```diff
@@ -1 +1,45 @@
-# write your code here
+import os
+import argparse
```
Oleksl888, line 2: here you can arrange import alphabetically

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        else:
+            data_input(file_name)
+
+
+def data_input(file_path: str) -> None:
+    count = 0
+    utc_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    mode = "a" if os.path.exists(file_path) else "w"
+    with open(file_path, mode) as file_time:
+        file_time.write(f"{utc_time}\n")
```
Oleksl888, line 32: use try except when working with file to catch IO related errors

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+            break
+        users_inputs.append(f"{count} {text}\n")
+    users_inputs.append("\n")
+    return users_inputs
+
+
+def data_input(file_path: str) -> None:
+    users_data = collect_data()
+    utc_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    with open(file_path, "a") as file_time:
```
Oleksl888, line 66: I saw you did file name validation but still it is a better to wrap manipulations with file into try - except since there are other problems that may appear when you work with file

PR #772, opened 2024-04-08
Solution

Nyxan:


app/create_file.py
```diff
@@ -1 +1,50 @@
...
+        dir_index = sys.argv.index("-d") + 1
+        file_index = sys.argv.index("-f") + 1
+        directory = os.path.join(*sys.argv[dir_index : file_index - 1])
+        filename = sys.argv[file_index]
+        content = []
+        while True:
+            line = input("Enter content line: ")
+            if line == "stop":
+                break
+            content.append(line)
```
YuriiKindrat, line 31: u could move this to the separate func e.g. write_content

app/create_file.py
```diff
@@ -1 +1,50 @@
...
+        create_file(directory, filename, content)
+    elif "-d" in sys.argv:
+        dir_index = sys.argv.index("-d") + 1
+        directory = os.path.join(*sys.argv[dir_index:])
+        os.makedirs(directory, exist_ok=True)
```
YuriiKindrat, line 36: same is here, 'create_dirs'

app/create_file.py
```diff
@@ -1 +1,50 @@
...
+    elif "-f" in sys.argv:
+        file_index = sys.argv.index("-f") + 1
+        filename = sys.argv[file_index]
+        content = []
+        while True:
+            line = input("Enter content line: ")
+            if line == "stop":
+                break
+            content.append(line)
+        create_file(os.getcwd(), filename, content)
```
YuriiKindrat, line 46: so there is won't be need to repeat urself

YuriiKindrat (changes requested):
try to use argparce to simplify ur code

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+    if os.path.exists(filepath):
+        with open(filepath, "a") as file:
+            file.write("\n\n" + timestamp + "\n")
+            for idx, line in enumerate(content_lines, start=1):
+                file.write(f"{idx} {line}\n")
+    else:
+        with open(filepath, "w") as file:
+            file.write(timestamp + "\n")
+            for idx, line in enumerate(content_lines, start=1):
+                file.write(f"{idx} {line}\n")
```
YuriiKindrat, line 20: don't repeat urself here, try to optimize ur solution 
the only thing that is changin is a mode and first line
u could move them e.g. to variables

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+import argparse
+from datetime import datetime
+
+
+def create_file(directory: str, filename: str, content_lines: list) -> None:
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+    filepath = os.path.join(directory, filename)
+
+    mode = "a" if os.path.exists(filepath) else "w"
```
Oleksl888, line 11: you can always use mode "a" here - if file not exists it creates new , otherwise adds to the previuos

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+    with open(filepath, mode) as file:
```
Oleksl888, line 12: How do you handle errors that may happen when you work with file?

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        file.write("\n\n" if mode == "a" else "" + timestamp + "\n")
+        for idx, line in enumerate(content_lines, start=1):
```
Oleksl888, line 14: you could add numbers and \n symbol to your content in get_content_lines method and then here call writelines() method

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+def get_content_lines() -> list:
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    content_lines = [timestamp + "\n"]
+    idx = 1
+    while True:
+        line = input("Enter content line: ")
+        if line.lower() == "stop":
+            break
+        content_lines.append(f"{idx} {line}\n")
+        idx += 1
```
Oleksl888, line 26: 1 last touch make an empty line if you append to file so your output looks like so:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content

File: app/create_file.py [hunk 9]
```diff
@@ -1 +1,51 @@
...
+
+    if args.directory:
+        directory_path = os.path.join(*args.directory)
+        os.makedirs(directory_path, exist_ok=True)
+        if args.file:
+            content_lines = get_content_lines()
+            create_file(directory_path, args.file, content_lines)
+    elif args.file:
+        content_lines = get_content_lines()
+        create_file(".", args.file, content_lines)
```
Oleksl888, line 47: can you simplify this so you can use only one if - statement. As you can see here you repeat the same logic twice:
```
content_lines = get_content_lines()
create_file(directory_path, args.file, content_lines)

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: app/create_file.py [hunk 10]
```diff
@@ -1 +1,49 @@
...
+    )
+    parser.add_argument("-d", "--directory", nargs="+", help="Directory path")
+    parser.add_argument("-f", "--file", help="File name")
+
+    args = parser.parse_args()
+
+    if args.directory:
+        directory_path = os.path.join(*args.directory)
+        os.makedirs(directory_path, exist_ok=True)
+    elif args.file:
```
Oleksl888, line 43: cannot be elif here - then if you pass dir1, dir2 it will create dirs but will never create a file. Plus be sure that you pass correct path to create_file func

Oleksl888 (changes requested):

app/create_file.py
Oleksl888: this hasnt been done - please check

Oleksl888 (approved):
Well done! But please check the last comment - it overwrites blank line on append

PR #774, opened 2024-04-08
Solution 1

semermaks:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+        while True:
+            line_count += 1
+            line_input = input("Enter content line: ")
+            if line_input == "stop":
+                break
+            file.write(f"{line_count} {line_input}\n")
+
+
+arguments = sys.argv
+create_file(arguments, create_folder(arguments))
```
YuriiKindrat, line 43: use name==main here, also it would be better to create a separate func e.g. 'main' to call this methods
also could u call create_folder func inside create_file method, so there is won't be need to pass path arg to create_file func

YuriiKindrat (changes requested):
also u could try to use argparce to simplify ur code

LiudmylaKulinchenko (approved):
Looks good ☀

PR #775, opened 2024-04-08
file from terminal

mikhailyemets:


app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+    with open(file_path, "a") as file:
+        file.write(f"{time} \n")
+    while 1:
+        line_number += 1
+        text_line = input("Enter content line: ")
+        if text_line.lower() == "stop":
+            break
+        with open(file_path, "a") as file:
+            file.write(f"{line_number} {text_line} \n")
```
YuriiKindrat, line 18: why do u need to use CM twice here ?
also there is no need to open file every time inside while loop, u could open the file only once outside the loop

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+        with open(file_path, "w") as file_time:
+            file_time.write(utc_time + "\n")
+        input_data(file_path)
+    elif "-d" in sys.argv and "-f" in sys.argv:
+        d_index = sys.argv.index("-d")
+        f_index = sys.argv.index("-f")
+        dir_path = os.path.join(*sys.argv[d_index + 1:f_index])
+        os.makedirs(dir_path, exist_ok=True)
+        file_path = os.path.join(dir_path, sys.argv[f_index + 1])
+        input_data(file_path)
```
YuriiKindrat, line 41: make ur code DRY here, try to use argparce, and move repeating logic to the separate functions

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+        os.makedirs(path_to_dir, exist_ok=True)
+    if file_path:
+        if dir_path:
+            path_to_file = os.path.join(*dir_path, file_path)
+            write_user_date(path_to_file)
+        else:
+            write_user_date(file_path)
+
+
+parser = argparse.ArgumentParser(description="Parser for args")
```
Oleksl888, line 40: this block can go into a function

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+parser.add_argument("-f", dest="file_path")
```
Oleksl888, line 41: you can add a message in `help` attr here that will be shown to user if he uses your comman incorrectly

app/create_file.py
```diff
@@ -1 +1,46 @@
-# write your code here
+import os
+from datetime import datetime
+import argparse
```
Oleksl888, line 3: you can sort in alphabetical order here so it's perfect :)

app/create_file.py
Oleksl888, line 40: Description can actually tell what exactly your parser does

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,60 @@
...
+    while 1:
+        user_input = input("Enter some message: ")
+        if user_input.strip().lower() == "stop":
+            break
+        user_input_data.append(f"{line_counter} {user_input}\n")
+        line_counter += 1
+    return user_input_data
+
+
+def write_user_date(file_name: str) -> None:
```
Oleksl888, line 19: probably you meant write_user_data :)

Oleksl888 (approved):

PR #776, opened 2024-04-08
Solution for task

oleksandra-shershen:


YuriiKindrat (approved):
GJ, also u could use argpace to simplify ur solution

PR #779, opened 2024-04-08
Solution

SashaKisliy:


app/create_file.py
```diff
@@ -1 +1,44 @@
...
+
+    if "-d" in args:
+        dir_index = args.index("-d") + 1
+        dir_names = args[dir_index:]
+        create_dir(dir_names)
+
+    if "-f" in args:
+        file_index = args.index("-f") + 1
+        file_name = args[file_index]
+        create_and_write_file(file_name)
```
YuriiKindrat, line 40: what if we will swap the flags ?
try to use Argparcer, it will much simplify ur code

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #781, opened 2024-04-08
'Solution'

kssshevchenko:


app/create_file.py
```diff
@@ -1 +1,65 @@
...
+            path_parts.append(arg)
+        elif file_flag and not file_name:
+            file_name = arg
+
+    path = os.path.join(*path_parts) if path_parts else ""
+
+    if dir_flag:
+        create_dir(path)
+    if file_flag:
+        create_file(path, file_name)
```
YuriiKindrat, line 61: it works fine, but try to use argparcer to simplify ur solution with arg parcing

YuriiKindrat (changes requested):

PR #782, opened 2024-04-08
Solution#1

VitalyBashkiser:


app/create_file.py
```diff
@@ -1 +1,49 @@
-# write your code here
+import os
+
+from typing import List
+
+import argparse
```
LiudmylaKulinchenko, line 5: Fix imports order

LiudmylaKulinchenko (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        create_file(args.f)
+
+
+def create_directories(dirs: List[str]) -> None:
+    path = os.path.join(*dirs)
+    try:
+        os.makedirs(path)
+        print(f"Directory created: {path}")
+    except FileExistsError:
+        print(f"Directory already exists: {path}")
```
YuriiKindrat, line 30: what if i'd like to write in existing one ?
use exist_ok arg for makedirs here

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+    path = os.path.join(*dirs)
+    try:
+        os.makedirs(path, exist_ok=True)
+        print(f"Directory created: {path}")
+    except FileExistsError:
+        print(f"Directory already exists: {path}")
+
+
+def create_file(filename: str) -> None:
+    with open(filename, "a") as file:
```
Oleksl888, line 34: how do you handle errors if bad file name is passed?

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        print("Enter content line (type 'stop' to finish):")
+        content = []
+        while True:
```
Oleksl888, line 37: Not a good design to open a while loop inside of an open file handle. Better put collecting input in a separate function and call it before you open file for writing.

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+            line = input()
+            if line.lower() == "stop":
+                break
+            content.append(line)
+
+        content = [f"{i + 1} {line}" for i, line in enumerate(content)]
+        file.write("\n".join(content))
+        print(f"File created: {filename}")
```
Oleksl888, line 45: this print is unnecessary and also misleading because you will not create a file if it already exists

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        print("Enter content line (type 'stop' to finish):")
+        content = []
+        while True:
+            line = input()
+            if line.lower() == "stop":
+                break
+            content.append(line)
+
+        content = [f"{i + 1} {line}" for i, line in enumerate(content)]
+        file.write("\n".join(content))
```
Oleksl888, line 44: > App should add current timestamp at the top and number lines. If file.txt already exists it should add content below:

this is not done

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,64 @@
...
+
+def create_file(filename: str) -> None:
+    content = collect_input_lines()
+
+    mode = "w" if not os.path.exists(filename) else "a"
+    with open(filename, mode) as file:
+        if mode == "a":
+            file.write("\n\n")
+
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
Oleksl888, line 54: no need to store timestamp in a  variable if you are going to use it only once

app/create_file.py
```diff
@@ -1 +1,64 @@
...
+        file.write(f"{timestamp}\n")
+
+        content = [f"{i + 1} {line}" for i, line in enumerate(content)]
+        file.write("\n".join(content))
```
Oleksl888, line 58: would be a better design not to call .write() 3 times. You could prepare data in your collect_input_files() func and then simply pass it to writelines() method here

app/create_file.py
```diff
@@ -1 +1,64 @@
...
+
+    mode = "w" if not os.path.exists(filename) else "a"
+    with open(filename, mode) as file:
+        if mode == "a":
+            file.write("\n\n")
+
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{timestamp}\n")
+
+        content = [f"{i + 1} {line}" for i, line in enumerate(content)]
```
Oleksl888, line 57: Also as per task requirements there should be a blank line between entries to file:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content

File: app/create_file.py [hunk 10]
```diff
@@ -1 +1,64 @@
...
+        if line.lower() == "stop":
+            break
+        content.append(line)
+    return content
+
+
+def create_file(filename: str) -> None:
+    content = collect_input_lines()
+
+    mode = "w" if not os.path.exists(filename) else "a"
```
Oleksl888, line 49: this is unnecessary - you can always open file in `a` mode

app/create_file.py
```diff
@@ -1 +1,64 @@
...
+    with open(filename, mode) as file:
```
Oleksl888, line 50: this block should be wrapped in try-except to catch all the related errors

Oleksl888 (approved):
I approve but please look through the comments carefully and make fixes to improve your code

PR #783, opened 2024-04-09
Solution

siaserb:


app/create_file.py
```diff
@@ -1 +1,38 @@
...
+
+    if "-d" in terminal_arguments and "-f" in terminal_arguments:
+        create_dirs(terminal_arguments[1], terminal_arguments[2])
+        file_path = new_dir_path = path.join(
+            getcwd(),
+            terminal_arguments[1],
+            terminal_arguments[2],
+            terminal_arguments[4]
+        )
+        filling_file_with_content(file_path)
```
YuriiKindrat, line 38: what if we will have more args then ur indexes ?
also it won't work if we will swap the args 
try to use Argparcer here, to simplify ur solution

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+            file.write(f"{line_index} {line}\n")
+
+
+if __name__ == "__main__":
+    terminal_arguments = argv[1:]
+    if terminal_arguments[0] == "-d" and "-f" not in terminal_arguments:
+        create_dirs(terminal_arguments[1], terminal_arguments[2])
+
+    if terminal_arguments[0] == "-f" and "-d" not in terminal_arguments:
+        filling_file_with_content(terminal_arguments[1])
```
YuriiKindrat, line 28: there is won't be need to write 'not' condition if u will place the last one to the first place )

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+                        help="Directory names",
+                        metavar=("dir_1", "dir_2"))
+
+    parser.add_argument("-f",
+                        "--file",
+                        help="File name")
+
+    args = parser.parse_args()
+
+    if args.dir and args.file:
```
Oleksl888, line 38: you can remove this if and move logic from line 40 to if on line 49 - this way you will not repeat your code

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+
+def create_dirs(dir_1: str, dir_2: str) -> None:
+    os.makedirs(os.path.join(os.getcwd(), dir_1, dir_2), exist_ok=True)
+
+
+def filling_file_with_content(file_name: str) -> None:
+    with open(file_name, "a") as file:
+        file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
+        line_index = 0
+        while True:
```
Oleksl888, line 14: 1.better separate logic of getting input from user into a separate function. 
2.No need to open a while a loop inside of open file - bad practice since the file may be blocked for writing indefinitely.
3.Wrap the part where you write to file in try-except to handle errors realted to file handling

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+
+
+def create_dirs(dir_1: str, dir_2: str) -> None:
+    os.makedirs(os.path.join(os.getcwd(), dir_1, dir_2), exist_ok=True)
+
+
+def get_user_input() -> str:
+    lines = []
+    line_index = 0
+    while True:
```
Oleksl888, line 13: should add blank line when appending to file like so:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content

File: app/create_file.py [hunk 6]
```diff
@@ -1 +1,62 @@
...
+
+    if args.dir:
+        create_dirs(args.dir[0], args.dir[1])
+        if args.file:
+            file_path = os.path.join(os.getcwd(),
+                                     args.dir[0],
+                                     args.dir[1],
+                                     args.file)
+            content = get_user_input()
+            write_content_to_file(file_path, content)
```
Oleksl888, line 55: could you rewrite this to avoid repetition? here and under line 56? it is basically same thing

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+            break
+        line_index += 1
+        lines.append(f"{line_index} {line}")
+    return "\n".join(lines)
+
+
+def write_content_to_file(file_name: str, content: str) -> None:
+    try:
+        with open(file_name, "a") as file:
+            file.write(f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}\n")
```
Oleksl888, line 25: you could collect all data including date in your `get_user_input` func and call writelines() method once here

Oleksl888 (approved):
Good job!. But please fix the latest comments

PR #786, opened 2024-04-09
Solution

messedxup:


app/create_file.py
```diff
@@ -1 +1,52 @@
...
+        try:
+            filename_index = sys.argv.index("-f") + 1
+            filename = sys.argv[filename_index]
+            print("Enter content line:")
+            while True:
+                line = input()
+                if line.lower() == "stop":
+                    break
+                content.append(line)
+            create_file(os.path.join(directory_path, filename), content)
```
YuriiKindrat, line 45: could u write separate function for writing content ?

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+def create_directory(path: str) -> None:
+    os.makedirs(path, exist_ok=True)
+
+
+def create_file(filename: str, content: list[str]) -> None:
+    with open(filename, "a") as file:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(timestamp + "\n")
+        for idx, line in enumerate(content, start=1):
+            file.write(f"{idx} {line}\n")
```
Oleksl888, line 15: you can prepare a list with data in your writing_content func and then simply call writelines with this list here.
Also remenber that if you append to file there must be blank line between different entries:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #787, opened 2024-04-09
Solution

Ruxxell54:


app/create_file.py
```diff
@@ -1 +1,41 @@
...
+            file.write(f"{line_number} {content_line}\n")
+            line_number += 1
+
+
+def main() -> None:
+    if "-d" in sys.argv and "-f" in sys.argv:
+        dir_index = sys.argv.index("-d") + 1
+        file_index = sys.argv.index("-f") + 1
+        directory = os.path.join(*sys.argv[dir_index:file_index])
+        filename = sys.argv[file_index]
```
YuriiKindrat, line 28: will this work correctly if we will swap the flags ?

YuriiKindrat (changes requested):
try to use Argparcer to simplify ur code

PR #788, opened 2024-04-09
Solution

Chub-andrew:


app/create_file.py
```diff
@@ -1 +1,38 @@
...
+            file.write(f"{count_number} {user_input} + \n")
+            count_number += 1
+
+
+def find_argument() -> None:
+    terminal_data = sys.argv
+    if "-d" in terminal_data and "-f" in terminal_data:
+        create_dir(terminal_data[2:-2])
+        file_path = os.path.join(*terminal_data[2:-2], terminal_data[-1])
+        create_file(file_path)
```
YuriiKindrat, line 30: will this work if we will swap the flags ?

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+        while True:
+            user_input = str(input(
+                'Write your string (or "stop" for ending): '))
+            if user_input.lower() == "stop":
+                break
+            file.write(f"{count_number} {user_input} + \n")
+            count_number += 1
+
+
+def find_argument() -> None:
```
YuriiKindrat, line 25: it is better to call it 'main'

YuriiKindrat (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        description="Create directories and file.")
+    parser.add_argument("-d", "--directory",
+                        nargs="*", help="Directory path to create")
+    parser.add_argument("-f", "--file",
+                        help="File to create")
+
+    args = parser.parse_args()
+
+    if args.directory:
+        create_dir(args.directory)
```
Oleksl888, line 43: your code fails when only directories are being passed:
`python create_file.py -d dir1 dir2`

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+
+
+def create_file(directory: str, name: str) -> None:
+    if directory:
+        name = os.path.join(*directory, name)
+    else:
+        name = os.path.join(name)
+
+    formatted_datetime = datetime.now().strftime("%Y-%m-%d, %H:%M:%S")
+    with open(name, "a") as file:
```
Oleksl888, line 20: code fails when bad file name is passed - use try except to catch errors here

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+import argparse
+import os
+
+
+from datetime import datetime
+
+
+def create_dir(dirs: list) -> None:
+    path = os.path.join(*dirs)
+    os.makedirs(path)
```
Oleksl888, line 10: this will fail if dirs already exist - use `exist_ok=True` flag

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+    if directory:
+        name = os.path.join(*directory, name)
+    else:
+        name = os.path.join(name)
+
+    formatted_datetime = datetime.now().strftime("%Y-%m-%d, %H:%M:%S")
+    with open(name, "a") as file:
+        count_number = 1
+        file.write(formatted_datetime + "\n")
+        while True:
```
Oleksl888, line 23: Will be better design to  collect input from user outside of open file in a separate function. Then pass collected data to write to file

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+def write_data_to_file(file_path: str, data: list) -> None:
+    formatted_datetime = datetime.now().strftime("%Y-%m-%d, %H:%M:%S")
+    with open(file_path, "a", encoding="utf-8") as file:
+        file.write(formatted_datetime + "\n")
+        for line in data:
+            file.write(line + "\n")
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(
```
Oleksl888, line 32: Your code doesn't work when you pass input without -d flag

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+        if user_input.lower() == "stop":
+            break
+        data.append(f"{count_number} {user_input}")
+        count_number += 1
+    return data
+
+
+def write_data_to_file(file_path: str, data: list) -> None:
+    formatted_datetime = datetime.now().strftime("%Y-%m-%d, %H:%M:%S")
+    with open(file_path, "a", encoding="utf-8") as file:
```
Oleksl888, line 25: as per requirements input should look like this when append, but your code doesn't add blank line:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content

Oleksl888 (approved):
Approved, but code is not working as expected 100%. Please fix

PR #790, opened 2024-04-11
Implement Create file from terminal task

bwaah1:


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+import sys
+from datetime import datetime
+
+
+def create_directory(path: str) -> None:
+    os.makedirs(path)
+
+
+def create_file(filename: str) -> None:
+    with open(filename, "a") as f:
```
Oleksl888, line 11: How do you handle errors that might occur when working with file? (bad file name, file missing etc)

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+        count = 1
+        f.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        while True:
```
Oleksl888, line 14: Not a good practice to open infinite loop inside open file. This may block file for writing indefinitely + consume extra system resources. Also would make sense to break down the code - collect input in a separate function and then give the prepared data to your create_file function

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+    index_f = sys.argv.index("-f")
+    directory_path = os.path.join(*sys.argv[2:index_f])
+    create_directory(directory_path)
+    create_file(os.path.join(directory_path, sys.argv[2]))
+
+elif "-d" in sys.argv:
+    create_directory(os.path.join(*sys.argv[2:]))
+
+elif "-f" in sys.argv:
+    create_file(sys.argv[2])
```
Oleksl888, line 33: you can simplify this by making only two if statements:
if -d in args
if -f in args
...
Also all this must be a part of your main() function and don't forget to add id __name__ = __main__

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+
+
+def create_file(
+        file_name: str,
+        content: list[str],
+        directory_name: str = ""
+) -> None:
+    path = os.path.join(*directory_name, file_name)
+    with open(path, "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
```
Oleksl888, line 18: you can prepare all data( including timestamp) in your get user input function and here just call method writelines on a list with data - that would be more efficient.
Also your output should look like so:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content

You are missing a blank line between data that is appended to same file.
Also you should wrap operations with writing to file in try-except to catch all possible errors. (for example now your code breaks if you pass `python create_file.py -d dir1/dir2 -f .`. But there can be other cases

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: app/create_file.py [hunk 5]
```diff
@@ -1 +1,68 @@
...
+import os
+import argparse
+
+
+def create_directory(directory_name: list[str]) -> None:
+    path = os.path.join(*directory_name)
+    try:
+        os.makedirs(path, exist_ok=True)
+    except OSError as e:
+        print(f"Error creating directory: {e}")
```
Oleksl888, line 11: if an error happens your program must exit - otherwise it will continue working with incorrect params

app/create_file.py
```diff
@@ -1 +1,68 @@
...
+) -> None:
+    path = os.path.join(*directory_name, file_name)
+    try:
+        with open(path, "a") as file:
+            file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+            for line in content:
+                file.write(f"{line}\n")
+            file.write("\n")  # Add a blank line between data
+    except OSError as e:
+        print(f"Error creating file: {e}")
```
Oleksl888, line 27: Same here - print an error message and exit program

app/create_file.py
```diff
@@ -1 +1,68 @@
...
+        content: list[str],
+        directory_name: str = ""
+) -> None:
+    path = os.path.join(*directory_name, file_name)
+    try:
+        with open(path, "a") as file:
+            file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+            for line in content:
+                file.write(f"{line}\n")
+            file.write("\n")  # Add a blank line between data
```
Oleksl888, line 25: No need for this comment - code is self-explanatory

app/create_file.py
```diff
@@ -1 +1,68 @@
...
+
+def create_file(
+        file_name: str,
+        content: list[str],
+        directory_name: str = ""
+) -> None:
+    path = os.path.join(*directory_name, file_name)
+    try:
+        with open(path, "a") as file:
+            file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
```
Oleksl888, line 22: So can you manage to move this to `get_user_input_content()` method?

Oleksl888 (changes requested):

MNDonut (changes requested):
Don't see any new commits that fix Oleksl888 comments

YuriiKindrat (approved):
GJ

PR #792, opened 2024-04-12
Solution

OttoLindholm:


app/create_file.py
```diff
@@ -1 +1,35 @@
...
+    if d_index and f_index:
+        return os.path.join(*argv[d_index + 1:f_index]), file_name
+    elif d_index:
+        return os.path.join(*argv[d_index + 1:]), file_name
+
+    return "", file_name
+
+
+if __name__ == "__main__":
+    dir_name, file_name = get_path(argv)
```
vasylhnatiuk, line 21: it would be better write all below code in function and only call function here

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #795, opened 2024-04-13
solution

haldaniko:


vsmutok (changes requested):
If `file.txt` already exists it should add content below

image

PR #796, opened 2024-04-13
Solution

mgoryn:


vsmutok (changes requested):
`python create_file.py -d dir1 dir2` - this command should not create a file

PR #797, opened 2024-04-14
solution

alonasorochynska:


YuriiKindrat (changes requested):
what about writing content inside file ?
also all checks chould pass, fix ur flake8 errors

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+        os.makedirs(path)
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        current_date = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{current_date}\n")
+        line = 1
+
+        while True:
```
Oleksl888, line 18: move this block to a separate function get_user_input or equivalent. Also calling infinite loop inside of an open file is not optimal. Collect input from user and then pass data to your create_file function.
Also the output of your file doesn't enumerate lines correctly:
```
1 Line1 ...
1 Line1 ...
1 Line1 ...
Line1 word shouldn't be there and line numbers should increase

File: app/create_file.py [hunk 2]
```diff
@@ -1 +1,41 @@
...
+
+
+def create_directory(dirs: list) -> None:
+    path = os.path.join(*dirs)
+    if not os.path.exists(path):
+        os.makedirs(path)
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
```
Oleksl888, line 13: how do you handle errors that may occur when working with file?

app/create_file.py
```diff
@@ -1 +1,41 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory(dirs: list) -> None:
+    path = os.path.join(*dirs)
```
Oleksl888, line 7: how do you handle case when user passes incorrect input here:
`create_file -d`

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+            file.write(f"{line} Line{line} {i}" + "\n")
+            line += 1
+
+        file.write("\n")
+
+
+def main() -> None:
+    args = sys.argv[1:]
+
+    if "-d" in args:
```
Oleksl888, line 38: The function still breaks when user passes incorrect input such as:
```
python create_file.py -d dir1 dirs -f
OR
python create_file.py -d

Use if statements or use try-except when processing user input so you don't get tracebacks

File: app/create_file.py [hunk 5]
```diff
@@ -1 +1,54 @@
...
+        dir_index = args.index("-d")
+        if "-f" in args:
+            file_index = args.index("-f")
+            dirs = args[dir_index + 1:file_index]
+        else:
+            dirs = args[dir_index + 1:]
+        create_directory(dirs)
+
+    if "-f" in args:
```
Oleksl888, line 47: When passed `python create_file.py -d dir1 dir2 -f file.txt` it creates file.txt outside dir1/dir2

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+    while True:
+        user_input = input("Write your content or enter 'stop' to exit: ")
+        if user_input == "stop":
+            break
+        user_data.append(user_input)
+    return user_data
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
```
Oleksl888, line 22: You should add try-except here - when working with file some things may go wrong

Oleksl888 (changes requested):

Oleksl888 (changes requested):
Please fix

app/create_file.py
```diff
@@ -1 +1,108 @@
...
+        user_data.append(user_input)
+    return user_data
+
+
+def choice_of_creation(dirs: list | None, file_name: str | None) -> None:
+    if dirs and file_name:  # dir and file
+        if not create_directory(dirs):
+            create_file(file_name, dirs)
+        else:
+            create_file(file_name, dirs)
```
YuriiKindrat, line 47: it seems to me, that u have the similar options in both conditions here

YuriiKindrat (changes requested):

Oleksl888 (approved):
I will approve but there is still things to fix:
1. Your program doesn't handle incorrect input - commands like `create_file -d` lead to tracebacks - you need to add extra checks for input or use try-except statements (preferable) to handle this.
2. You must wrap writing to file in try-except block - to handle all related errors.
3. Output looks like so: `1 Line1 ...` you don't need to add `Line1` etc to the output since you enumerate lines already. It doesn't add anything to the code 
4. You could simplify your code overall - you have too many if statements in your `main`, '`choice of creation`, `check_presence..` functions. For example you could make a function 
```
def create_directories(args: list) -> str:
    try:
        slice_end_index = args.index("-f")
    except ValueError:
        slice_end_index = len(args) + 1

    dirs_list = args[args.index("-d") + 1: slice_end_index]
    dir_path = os.path.join(*dirs_list)
    os.makedirs(dir_path, exist_ok=True)
    return dir_path

and then call it in your main in try-except block:
```
    if args[0] == "-d":
        try:
            dir_path = create_directories(args)
        except TypeError:
            print("Cannot create directory! Usage: python create_file.py -d dir1 dir2 -f your_file_name.txt")
            exit()

PR #798, opened 2024-04-15
Solution

sixsixsixsamurai:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+import datetime
+
+
+def create_file(file_name: str) -> None:
+    with open(file_name, "a") as file:
+        file.write(
+            str(datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S\n")))
+        line_number = 1
+        while True:
+            input_content = input("Enter content line:")
```
Dimosphen1, line 12: Add a space after `:`
```suggestion
            input_content = input("Enter content line: ")
```

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        return
+
+    if "-f" in cmd_text:
+        index_f = cmd_text.index("-f")
+        create_file(cmd_text[index_f + 1])
+        return
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 48: There are several issues with the code:
- There should be no blank lines at the end of the file;
- Delete `if "-d" in cmd_text and "-f" in cmd_text:` condition, you should leave only the last two if conditions without returns,so that in case `-d` flag is present you create directories and if `-f` flag is present as well, you create file as well

Dimosphen1 (changes requested):

pavlopro (changes requested):
There should be a blank line between different input sessions

pavlopro (changes requested):
create_file.py -f test.txt -d dir1 dir2

This produces an error. Please check

PR #803, opened 2024-04-16
Solution

mvsydoruk:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    file_name = args[-1]
+
+if dir_list and file_name:
+    create_directory(dir_list)
+    file_path = os.path.join(*dir_list, file_name)
+    create_file(file_path)
+elif not dir_list and file_name:
+    create_file(file_name)
+elif dir_list and not file_name:
+    create_directory(dir_list)
```
Dimosphen1, line 43: Put this part inside of the `if __name__ == "__main__":` block

Dimosphen1, line 43: Use only two if conditions here so that in case `dir_list` is present you create directories and update `file_name` and in case file is present you create the corresponding file

Dimosphen1, line 43: There should be no blank lines at the end of the file, the file content should be the same as one in the task description

Dimosphen1 (changes requested):

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #805, opened 2024-04-17
Solution

d-kaliuta:


EdAlekseiev (changes requested):
You should split your code to separate function

EdAlekseiev (approved):

PR #813, opened 2024-04-26
'Solution'

Vitalii-Haponiuk:


pavlopro (changes requested):
Check the result of `python create_file.py -f file -d dir3 dir4 dir5`
It doesn't work correctly

PR #818, opened 2024-05-06
'sol'

liudmylaperchevska:


app/create_file.py
```diff
@@ -1 +1,30 @@
...
+
+        while True:
+            content = input("Enter content line: ")
+            if content.lower() == "stop":
+                break
+            new_file.write(content + "\n")
+
+else:
+    file_path = os.path.join(*arguments[2:])
+    os.makedirs(file_path, exist_ok=True)
```
Dimosphen1, line 30: - There should not be blank line at the end of the file, file should be the same as one from the example;
- You should create line enumeration in the file;
- File should not be erased after trying to append to the end new content, it should be added with one blank line from the previous input;
- Divide logic between functions, then group them inside of the `main` function and call this main() in the if `__ name__ == __"main__"` block.

app/create_file.py
```diff
@@ -1 +1,30 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+print("Argument List:", sys.argv)
```
Dimosphen1, line 6: Delete redundant prints, there is no need for it

Dimosphen1 (changes requested):

PR #825, opened 2024-05-09
Solution

Roman-Sokolov-V:


pavlopro (changes requested):
Check the example from the task description. Each line must be numbered.
For example:

2022-02-01 14:46:01
1 Line1 content
2 Line2 content
3 Line3 content

pavlopro (approved):
Good Job!

PR #832, opened 2024-05-22
create file from terminal sol

TenbReala:


app/create_file.py
```diff
@@ -1 +1,64 @@
-# write your code here
+import os
+import argparse
+from datetime import datetime
+
+
+def create_directory(dir_path: str) -> None:
```
Dimosphen1, line 6: There is no need for this method, you could call `os.makedirs` directly where needed

app/create_file.py
```diff
@@ -1 +1,64 @@
...
+        if args.file_name:
+            file_path = os.path.join(directory, args.file_name)
+            create_file(file_path)
+    else:
+        file_path = os.path.join(args.file_name)
+        create_file(file_path)
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 64: - There should not be `Enter content (use 'stop' to finish editing):` text at the beginning of the code execution, follow the description of the task precisely;
- There should be no blank line at the end of a newly created file;
- There should be one blank line between different input sessions to the file, not two.

Dimosphen1 (changes requested):

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+
+        line_count = 0
+        while True:
+            line_count += 1
+            line = input("Enter content line: ")
+            if line.lower() == "stop":
+                break
+            file.write(f"\n{line_count} {line}")
+
+
```
pavlopro, line 24: There should be blank line between different input sessions to the file.

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #835, opened 2024-05-24
Solution

MatsalakO:


pavlopro (changes requested):
There should be current timestamp and one blank line between different input sessions (check example in task description)

It doesn't work correct with next command:

pavlopro (changes requested):
An error occurred when:
python create_file.py -f file.txt

pavlopro (changes requested):
Good Job!

pavlopro (approved):
Good Job!

PR #838, opened 2024-05-25
added solution to create file from terminal

AlexRozv:


LiudmylaKulinchenko (approved):
My suggestions:
1. Consider split on functions (for example, add_content ect.)
2. You can also use argparse module here — it would be easier, more readable, and you won't have to reinvent the wheel

PR #840, opened 2024-05-26
Solution

mariia-sergeevna:


mariia-sergeevna:
> App should add current timestamp at the top **and number lines**. Add numbers lines image

fixed

PR #842, opened 2024-05-29
solution

AndrewwSSS:


app/create_file.py
```diff
@@ -1 +1,50 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+# get information from console args
+i = 1
```
DevSpaciX, line 6: Why you decided to call this viriable i ? Maybe let`s use more descriptive naming ?

app/create_file.py
```diff
@@ -1 +1,50 @@
...
+path = ""
+d_flag = False
+f_flag = False
+file_name = ""
```
DevSpaciX, line 10: Look you can use sys.argv to check if flag was in the terminal and split every single func by this condition 
For exmpl 
```
if "-d" in sys.argv:
    create_dirs(sys.argv)
```

`create_dirs` - it will be your func

DevSpaciX (changes requested):
Split your code by functions that will meet the flags conditions

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+        if len(sys.argv) < flag_idx + 1:
+            raise ValueError("-f flag requires one parameter")
+        result_path = os.path.join(result_path, sys.argv[flag_idx + 1])
+    else:
+        raise ValueError("There is not -f flag")
+    return result_path
+
+
+# get information from console args
+path = get_path_from_args()
```
Oleg-s-s, line 31: You can theoretically put that function into context manager and there will be no need in additional value

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+with open(path, "w") as f:
+    f.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+    line_counter = 1
+    data = ""
+    while True:
+        data = input("Enter content line: ")
+        if data == "stop":
+            break
+        f.write(f"{line_counter} {data}\n")
+        line_counter += 1
```
Oleg-s-s, line 44: Looks like a file creation function to me

Oleg-s-s (changes requested):
Don't forget to follow basic procedural principles and use `if name == main` after you will implement changes requested

PR #845, opened 2024-06-04
py-create-file-from-terminal v1

BornToLivee:
my brain died after this task :)


DevSpaciX (approved):
By the way you did mega nice !)

PR #847, opened 2024-06-05
Solution

rinarice:


vsmutok (changes requested):
```bash
python create_file.py -d dir1 dir2 dir3 
```
after this command I am prompted to enter data into a file

image

app/create_file.py
```diff
@@ -1 +1,70 @@
...
+def main() -> None:
+    args = sys.argv[1:]
+    path_parts = []
+    file_name = None
+
+    if "-d" in args and "-f" in args:
+        dir_index = args.index("-d")
+        file_index = args.index("-f")
+        file_name = args[file_index + 1]
+        path_parts = args[dir_index + 1:file_index]
```
EdAlekseiev, line 57: It doesn't work for `python create_file.py -f file.txt -d dir1 dir2`

EdAlekseiev (changes requested):

app/create_file.py
rinarice: Hi Ed, thanks for your feedback! 

Can you guide me please how should it work? Should it create a file.txt in dir1/dir2 when the prompt with -f flag goes first and only then directories are mentioned?

EdAlekseiev (approved):

PR #868, opened 2024-06-10
solution

DimaYankovskiy:


app/create_file.py
```diff
@@ -1 +1,33 @@
...
+if "-d" in sys.argv:
+    if "-f" in sys.argv:
+        end_index = sys.argv.index("-f")
+    else:
+        end_index = len(sys.argv)
+    path = "/".join(sys.argv[sys.argv.index("-d") + 1:end_index])
+
+    os.makedirs(path, exist_ok=True)
+
+    path += "/"
```
DevSpaciX, line 17: Put your logic into a functions

app/create_file.py
```diff
@@ -1 +1,33 @@
...
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"\n{timestamp}\n")
+
+        numerator = 1
+        while True:
+            line = input("Enter content line: ")
+            if line.lower() == "stop":
+                break
+            file.write(f"{numerator} {line}\n")
+            numerator += 1
```
DevSpaciX, line 33: And this should be in func too . Is is bad practice to write code like a whole file

DevSpaciX (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+from datetime import datetime
+
+
+def resolve_directory() -> str:
+    if "-f" in sys.argv:
+        end_index = sys.argv.index("-f")
+    else:
+        end_index = len(sys.argv)
+
+    directory_path = "/".join(sys.argv[sys.argv.index("-d") + 1:end_index])
```
DevSpaciX, line 12: Try to use os.path.join

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #869, opened 2024-06-10
Develop

MaglorFeanorson:


app/create_file.py
```diff
@@ -1 +1,29 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+path = ""
+if "-d" in sys.argv:
+    if "-f" in sys.argv:
```
DevSpaciX, line 8: May be better use `and` operator to make your code less nested ?)

app/create_file.py
```diff
@@ -1 +1,29 @@
...
+        end_index = sys.argv.index("-f")
+    else:
+        end_index = len(sys.argv)
+    path = "/".join(sys.argv[sys.argv.index("-d") + 1:end_index])
```
DevSpaciX, line 12: You can try to use `os.path.join` here

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #870, opened 2024-06-10
sol1

Unlie9:


app\file.txt
```diff
@@ -0,0 +1,4 @@
+23423
+42342
+2423
+break
```
DevSpaciX, line 4: Delete redundant  files from commit

app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import os
+import sys
+
+
+def make_path() -> str:
+    file_path = os.getcwd()
+    file_name = "file.txt"
+
+    i = 1
```
DevSpaciX, line 9: `i` it is not very good practive to name viriables : ( Fix it

DevSpaciX (changes requested):

app\file.txt
DevSpaciX: Look how to remove files from commit  and remove it pls )

DevSpaciX (changes requested):

DevSpaciX (approved):
Mega super good !

PR #875, opened 2024-06-16
solution

Fedot0v:


app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+def create_file(file_name: str, path: str = ".") -> None:
+    file_path = os.path.join(path, file_name)
+    with open(file_path, "a") as file:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{timestamp}\n")
+
+        number_of_line = 1
+        while True:
+            content_line = input(f"Enter content line {number_of_line}: ")
```
Dimosphen1, line 21: There is no need to add line number in the input section, follow task description

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+        path = create_directory(command[2:-2])
+        create_file(command[-1], path)
+    elif "-d" in command:
+        create_directory(command[2:])
+    elif "-f" in command:
+        create_file(command[-1])
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 40: - You should add one blank line between different input sessions while appending to an existing file;
- There should be no blank lines at the end of the newly created or edited file according to the task example.

app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory(args: list) -> str | bytes:
+    path = os.path.join(*args)
+    if not os.path.exists(path):
+        os.makedirs(path)
```
Dimosphen1, line 9: You could use `exist_ok=True` flag to avoid using this if condition

Dimosphen1 (changes requested):

pavlopro (changes requested):
python create_file.py -f file.txt -d dir3

This command doesn't work correctly. Check also another possible variations.

Created file contains blank line at the start

pavlopro (changes requested):
python create_file.py -f file.txt -d dir3

This command doesn't work correctly
It creates a directory 'file.txt' with file named 'dir3'

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+            content_line = input("Enter content line: ")
+            if content_line.lower() == "stop":
+                break
+            file.write(f"\n{number_of_line} {content_line}")
+            number_of_line += 1
+
+
+def main() -> None:
+    command = sys.argv
+    print(command)
```
vsmutok, line 39: remove this `print`

app/create_file.py
```diff
@@ -1 +1,57 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory(path_parts: list) -> str | bytes:
```
vsmutok, line 6:
```suggestion
def create_directory(path_parts: list) -> str:
```
function always returns a string

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+
+
+def create_file(
+        file_name: str,
+        path: str | os.PathLike | bytes = ".",
+        mode: str = "w"
+) -> None:
+    file_path = os.path.join(path, file_name)
+    if os.path.exists(file_path):
+        mode = "a"
```
vsmutok, line 19: I think this is unnecessary, it is better to use `mode = "a"` by default

app/create_file.py
```diff
@@ -1 +1,57 @@
...
+    path = os.path.join(os.getcwd(), *path_parts)
+    os.makedirs(path, exist_ok=True)
+    return path
+
+
+def create_file(
+        file_name: str,
+        path: str | os.PathLike | bytes = ".",
+        mode: str = "w"
+) -> None:
```
vsmutok, line 16:
```suggestion
def create_file(file_name: str, path: str, mode: str = "w") -> None:
```

vsmutok (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #879, opened 2024-06-20
Solution

IraMladanovych02:


app/create_file.py
```diff
@@ -1 +1,67 @@
...
+
+    if file_name:
+        file_path = os.path.join(directory, file_name)
+        content_lines = get_file_content()
+        write_to_file(file_path, content_lines)
+        print(f"File created/updated at: {file_path}")
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 67: Do not leave one blank line at the beginning and at the ending of the file, file content should be exactly the same as one from the task description.

app/create_file.py
```diff
@@ -1 +1,67 @@
...
+    if dir_path:
+        directory = create_directories(dir_path)
+    else:
+        directory = os.getcwd()
+
+    if file_name:
+        file_path = os.path.join(directory, file_name)
+        content_lines = get_file_content()
+        write_to_file(file_path, content_lines)
+        print(f"File created/updated at: {file_path}")
```
Dimosphen1, line 63: Remove this `print` statement, there is no need in it

Dimosphen1 (changes requested):

PR #886, opened 2024-06-27
done

Ppllupold:


vsmutok (changes requested):
when I use this command, folders not created 
```
```

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+            text = input("Enter content line: ")
+            if text == "stop":
+                file.write("\n")
+                break
+            file.write(f"{line_num} {text}\n")
+            line_num += 1
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 48: Do not leave two blank lines at the end of the file after creating or updating it. There should be no blank lines at the end of the file, the file content should be the same as the one from the task description

Dimosphen1 (changes requested):

PR #887, opened 2024-06-28
+

AndriZhok:


vsmutok (changes requested):
this command does not work
```
```
image

AndriZhok:
there, on the contrary, it is necessary to enter. If you enter first -d and then -f, then everything is fine

PR #894, opened 2024-07-07
"sol"

arsenmarkotskyi:


app/create_file.py
```diff
@@ -1 +1,55 @@
...
+import sys
+import os
+from datetime import datetime
+
+
+def create_file(path: str, file_name: str) -> None:
+    file_path = os.path.join(path, file_name)
+
+    if not os.path.exists(path):
+        os.makedirs(path)
```
EdAlekseiev, line 10: You can simplify your code with using `exist_ok =True`
```suggestion
    os.makedirs(path, exist_ok =True)
```

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+
+    mode = "a" if os.path.exists(path) else "w"
```
EdAlekseiev, line 12: You can use "a" always
"a" - Append - Opens a file for appending, creates the file if it does not exist
```suggestion
    mode = "a"
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #895, opened 2024-07-08
Solution

EugeneDolinskiy:


app/create_file.py
```diff
@@ -1 +1,79 @@
...
+    if "-f" in args:
+        f_index = args.index("-f")
+        file_name = args[f_index + 1]
+
+    return directories, file_name
+
+
+def create_directories(
+        directories: List[str]
+) -> LiteralString | str | bytes:
```
pavlopro, line 26: Please simplify this by removing `LiteralString` to be able to test it with py3.10

pavlopro (changes requested):

pavlopro (changes requested):
Check this `python3 create_file.py -f test.txt -d dir1 dir2` it doesn't work

Currently you program adds blank line after the session ends
But it shoud add blank line only when session starts

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+    os.makedirs(path_to_file, exist_ok=True)
+
+
+def create_new_file(new_file: str) -> None:
+    with open(new_file, "a") as result_file:
+        creation_time = datetime.datetime.now()
+        counter = 1
+        result_file.write(f"{creation_time.strftime('%Y-%m-%d %H:%M:%S')}\n")
+        while True:
+            new_text_line = input("Enter additional "
```
pavlopro, line 17: Please correct the text to "Enter content line:"

There shouldn't be any blank line in the end of input session

pavlopro (changes requested):
This doesn't work correct

python create_file.py -f test.txt -d dir1 dir2

app/create_file.py
EugeneDolinskiy: I don't really understand how to make it so that an empty line is not added at the end of the session, but is added at the start of a new session. Could you tell me how to do this?

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+        line_counter = 1
+        while (line := input("Enter content line: ")) != "stop":
+            file.write(f"{line_counter} {line}\n")
+            line_counter += 1
+        file.write("\n")
+
+
+if __name__ == "__main__":
+    main(sys.argv[1:])
```
Dimosphen1, line 40: Please, make sure that the file content is the same as in the task description:
- There are no blank lines at the end of the file;
- There is one blank line between different input sessions.

Dimosphen1 (changes requested):

PR #898, opened 2024-07-09
solution

midetc:


vasylhnatiuk (changes requested):
image
Your code doesn't work correctly. After the second running, text should be added to the preceding; previous text shouldn't be deleted! and also add numeration lines.

YukihiroSM (approved):
Looks good to me! Nice job!

PR #904, opened 2024-07-15
Develop

aniashev:


file.txt
EdAlekseiev, line 1: Delete your test file from repositry, please

EdAlekseiev (changes requested):

vsmutok (changes requested):
This command should create folders, and inside them a file
```
```

vasylhnatiuk (approved):

PR #905, opened 2024-07-17
Try Solution

D0mln0:


vsmutok (changes requested):
```
python create_file.py -d dir1 dir2 dir3 -f file.txt
```
this command should create file inside a directory

vasylhnatiuk (approved):

PR #908, opened 2024-07-21
File create

itpsychologist:


pavlopro (changes requested):
This doesn't work correcly

Also, there should be an empty line between different input sessions

PR #912, opened 2024-07-25
Implemented my miserable solution

dimak20:
It works (also with extra conditions including the case when user writes -f key before -d key), but looks so ugly. Should I separate module's logic into functions and then call all of them?

### DIFF
--- a/app/create_file.py
+++ b/app/create_file.py
@@ -1,1 +1,37 @@
-# write your code here
+from datetime import datetime
+import sys
+import os
+
+
+data = sys.argv[1:]
+f_index = data.index("-f")
+
+if "-d" in data:
+    d_index = data.index("-d")
+    if f_index > d_index:
+        d_data = data[d_index + 1: f_index]
+        path_dirs = os.path.join(*d_data)
+        file_path = os.path.join(path_dirs, data[f_index + 1])
+    else:
+        d_data = data[d_index + 1:]
+        path_dirs = os.path.join(*d_data)
+        file_path = os.path.join(path_dirs, *data[f_index + 1: d_index])
+
+    os.makedirs(path_dirs, exist_ok=True)
+
+else:
+    file_path = os.path.join(data[f_index + 1])
+
+current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+counter = 1
+info_from_user = f"{current_time}\n"
+
+while True:
+    line_from_user = input("Enter content line: ")
+    if line_from_user == "stop":
+        break
+    info_from_user += f"{counter} {line_from_user}\n"
+    counter += 1
+
+with open(file_path, "w") as file:
+    file.write(info_from_user)
--- a/app/create_file.py
+++ b/app/create_file.py
@@ -33,5 +33,5 @@ while True:
     info_from_user += f"{counter} {line_from_user}\n"
     counter += 1
 
-with open(file_path, "w") as file:
+with open(file_path, "a") as file:
     file.write(info_from_user)


app/create_file.py

app/create_file.py
```diff
@@ -1 +1,35 @@
...
+    file_path = os.path.join(data[f_index + 1])
+
+current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+counter = 1
+info_from_user = f"{current_time}\n"
+
+while True:
+    line_from_user = input("Enter content line: ")
+    if line_from_user == "stop":
+        break
```
oliasymak24, line 30: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+            break
+        info_from_user += f"{counter} {line_from_user}\n"
+        counter += 1
+
+    with open(path, "a") as file:
+        file.write(info_from_user)
+
+
+def create_file() -> None:
+    sys_data = sys.argv[1:]
```
Oleksl888, line 35: you can put logic of parsing arguments in a separate function

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    f_index = sys_data.index("-f")
+    if "-d" in sys_data:
```
Oleksl888, line 37: Does your code handle invalid commands such as `python create_file.py -d -f file.txt`

app/create_file.py
```diff
@@ -1 +1,45 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+from typing import LiteralString
+
+
+def create_dirs(sys_data: list, f_index: int) -> LiteralString | str | bytes:
```
Oleksl888, line 7: when does this function return `bytes` object? Also why use Literalstring here?

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    os.makedirs(os.path.join(*d_data), exist_ok=True)
+    return os.path.join(*d_data, sys_data[f_index + 1])
+
+
+def write_data(path: LiteralString | str | bytes) -> None:
+    current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    counter = 1
+    info_from_user = f"{current_time}\n"
+
+    while True:
```
Oleksl888, line 22: it would make sense to put collecting user input in a separate function

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        line_from_user = input("Enter content line: ")
+        if line_from_user == "stop":
+            info_from_user += "\n"
+            break
+        info_from_user += f"{counter} {line_from_user}\n"
```
Oleksl888, line 27: I think it will be more memory efficient to use list and append lines to it rather then creating new string every time you call `+=` on a string. You can pass list to writelines method afterwards

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        counter += 1
+
+    with open(path, "a") as file:
```
Oleksl888, line 30: wrap this block in try except to catch file-related errors

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,83 @@
...
+    f_flag = None
+    d_flag = None
+    file_name = ""
+    dir_path = ""
+    if "-f" in arg:
+        f_flag = True
+        f_index = arg.index("-f")
+    if "-d" in arg:
+        d_flag = True
+        d_index = arg.index("-d")
```
oliasymak24, line 48: try to use argparse, it will look much more better

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+    return "".join(info)
+
+
+def write_data(file_name: str, path: list = None) -> bool:
+    file_path = os.path.join(*path, file_name) if path else file_name
+    data_from_user = collect_data()
+    try:
+        with open(file_path, "a") as file:
+            file.write(data_from_user)
+    except (PermissionError, OSError) as e:
```
Oleksl888, line 31: PermissionError is actually a subclass of OSError

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+        print(f"Cannot open/create file: {file}. Error {e}")
+        return False
+    else:
+        return True
+
+
+def parse_argv() -> tuple | None:
```
Oleksl888, line 38: it actually always return a tuple

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+
+
+def create_file() -> None:
+    file, dirs = parse_argv()
+    if file:
+        if dirs:
+            create_dirs(dirs)
+        is_success = write_data(file, dirs)
+        if is_success:
+            print("User data has written")
```
Oleksl888, line 55: You don't really need these prints - it's ok for debugging but you should remove them when you push the code to remote.
In reality print statements slow down your system since they take computation power, so web apps use `logging` instead and logs are usually set to errors/warnings only, so when your program behaves as expected you don't log anything.

app/create_file.py
```diff
@@ -1 +1,66 @@
...
+    else:
+        return True
+
+
+def parse_argv() -> tuple | None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-f", "--file_name", type=str)
+    parser.add_argument("-d", "--dir_path", type=str, nargs="+")
+    sys_args = parser.parse_args()
+    file_name = sys_args.file_name if sys_args.file_name else None
```
Oleksl888, line 43: you don't really need these ternary statements - argparser should return None by default if no args are passed

Oleksl888 (changes requested):
Overall good job - just need to do some cosmetic changes

Oleksl888 (approved):
Well done! :muscle:

PR #916, opened 2024-07-28
solution

Morphin20th:


vsmutok (changes requested):
`python create_file.py -f file.txt -d dir1 dir2 dir3` - if i run this command a second time it throws an error

image

app/create_file.py
```diff
@@ -1 +1,28 @@
-# write your code here
+import os
+import argparse
+from datetime import datetime
+
```
pavlopro, line 4: 1. Lines should be enumerated (check example in task description)
2. There shouldn't be any blank lines in the end of file

`python3 create_file.py -d dir`
Gives me an error

pavlopro (changes requested):

EdAlekseiev (approved):

PR #917, opened 2024-07-29
Solution

OleksiukStepan:


app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    directory = ""
+
+    if args[0] == "-d":
+        directory = args[1:]
+        if "-f" in args:
+            directory = directory[:args.index("-f") - 1]
+            args = args[args.index("-f"):]
+
+        try:
+            os.makedirs("/".join(directory))
```
Oleksl888, line 17: you should use os,path,join instead to correctly build path on all systems

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+        except FileExistsError:
+            print("Directory already exist")
```
Oleksl888, line 19: shouldnt be an exception here - if file exists, you should write to the same file

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+import os.path
+import sys
+from datetime import datetime
+
+
+def create_file() -> None:
+    args = sys.argv[1:]
+    directory = ""
+
+    if args[0] == "-d":
```
Oleksl888, line 10: it would be a better design to further separate code in functions. Make a function parse_arguments and put all the related logic there

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+
+    if args[0] == "-f":
+        try:
+            _make_file(directory, args[1])
+        except FileNotFoundError:
+            print("File not found")
+
+
+def _make_file(directory: list[str], filename: str) -> None:
+    mode = "w"
```
Oleksl888, line 29: you can always write to file in "a" mode

app/create_file.py
```diff
@@ -1 +1,43 @@
...
+    if os.path.exists(filename):
+        mode = "a"
+    with open(os.path.join(*directory, filename), mode) as file:
+        current_date = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
Oleksl888, line 33: you can put collecting data in a separate function - e.g. `get_user_data` where you will prepare data for writing. 
Also I would suggest collecting data in an infinite loop outside an open file handle - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #918, opened 2024-07-30
"Solution py-create-file-from-terminal"

Olga-Gynguliak:


vsmutok (changes requested):
this command not work 
`python create_file.py -f file.txt -d dir1 dir2 dir3`

PR #924, opened 2024-08-04
Solution

DSahalatyi:


app/create_file.py
```diff
@@ -1 +1,47 @@
...
+import os
+import sys
+from datetime import datetime
+
+
+def path_handler(file_name: str | None, path: str | None) -> None:
+    if path and not os.path.exists(path):
+        os.makedirs(path)
+    else:
+        path = ""
```
Oleksl888, line 10: this logic  is incorrect - if you give your script an existing path it will rewrite it to empty path.

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+
+    if file_name:
+        with open(os.path.join(path, file_name), "a") as file:
+            lines = []
+            n_line = 1
+
+            while True:
```
Oleksl888, line 17: Collecting user input should be placed in a separate function. Also I would suggest collecting data in an infinite loop outside an open file handle - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+                if line == "stop":
+                    break
+                lines.append(f"{n_line} {line}\n")
+                n_line += 1
+
+            current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+            file.write(f"{current_time}\n")
+            file.writelines(lines)
+            file.write("\n")
```
Oleksl888, line 28: maybe you could optimize this by getting all data in a list and then writing it all at once with writelines method?

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+if __name__ == "__main__":
+    args = sys.argv
+    file_name = None
+    path = None
+
+    if "-d" in args:
+        path = ""
+        i = args.index("-d") + 1
+
+        while i < len(args) and not args[i].startswith("-"):
```
Oleksl888, line 40: you can put it in a separate function to make your code more organized

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+
+
+def path_handler(file_name: str | None, path: str | None) -> None:
+    if path and not os.path.exists(path):
+        os.makedirs(path)
+    else:
+        path = ""
+
+    if file_name:
+        with open(os.path.join(path, file_name), "a") as file:
```
Oleksl888, line 13: would be nice to wrap writing to file with try except to catch all possible errors when working with file

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #926, opened 2024-08-06
solution

BastovOleksandr:


app/create_file.py
```diff
@@ -1 +1,67 @@
...
+        return command[command.index(flag) + 1]
+
+
+def process_flag(path_or_file_name: str, flag: str) -> None:
+    if path_or_file_name:
+        path = os.path.normpath(os.getcwd() + os.sep + path_or_file_name)
+
+        if not os.path.exists(path):
+            if flag == D_FLAG:
+                os.makedirs(path)
```
oliasymak24, line 32: u can use makedirs argument "exist_ok" here

app/create_file.py
```diff
@@ -1 +1,67 @@
...
+    all_lines = ""
+    line_counter = 1
+
+    with (open(path, "a+") as file):
+        while True:
+            input_line = input(
+                'Enter content line or "stop" to break input: '
+            )
+            if input_line == "stop":
+                break
```
oliasymak24, line 51: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

oliasymak24 (changes requested):

PR #927, opened 2024-08-07
'Solution'

oleksashcherbakov:


app/create_file.py
```diff
@@ -1 +1,56 @@
...
+        new_string = input("Enter new line of content: ")
+        if new_string != "stop":
+            lines_arr.append(new_string)
+        else:
+            break
+    with open(file_name, "a") as file:
+        current_time = datetime.datetime.now()
+        file.write(current_time.strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        for line in lines_arr:
+            file.write(f"{line}\n")
```
Dimosphen1, line 56: Make sure that the execution call is present in the file in the `if __name__ == '__main__':` block and it works correctly in scenarios that are mentioned in the task. File content should be exactly the same as one from the task description

Dimosphen1 (changes requested):

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+        for index, line in enumerate(lines_arr):
+            file.write(f"{index + 1} {line}\n")
+
+
+def check_directories(taken_path: str) -> bool:
+    return os.path.exists(taken_path)
+
+
+if __name__ == "__main__":
+    create_file(sys.argv)
```
Dimosphen1, line 56: - Command  `python create_file.py -f file.txt -d d1 d2 d3` is not working;
- Make sure that there are no blank lines at the end of the newly created or appended file;
- Make sure that there is a blank line between different input sessions to the same file.

Dimosphen1 (changes requested):

app/create_file.py
oleksashcherbakov: have to be OK

app/create_file.py
```diff
@@ -1 +1,69 @@
...
```
Dimosphen1, line 69: There is still one blank line at the end of the file. Make sure that it is absent after new file creation or update

Dimosphen1 (changes requested):
Several minor changes were requested

app/create_file.py
oleksashcherbakov: done

app/create_file.py
```diff
@@ -1 +1,74 @@
-# write your code here
+import os
+import datetime
+import sys
+
+
+def create_file(arguments: list) -> None:
+    path = define_directories(arguments)
+    if not check_directories(path):
```
pavlopro, line 8: Simplify to remove this `check_directories `method
os.makedirs(path, exist_ok=True) already implements this check

pavlopro (changes requested):

app/create_file.py
oleksashcherbakov: deleted

pavlopro (approved):
GJ!

PR #928, opened 2024-08-07
Solution

dartomOOv:


app/create_file.py
```diff
@@ -1 +1,73 @@
...
+    try:
+        os.makedirs(path_to_implement)
+    except OSError as error_info:
+        print(error_info)
+
+
+full_command = sys.argv
+command, *another_part_of_command = full_command[1:]
+
+if command == "-f":
```
Oleksl888, line 52: i recommend to put parsing arguments in a separate function

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+        path = os.path.join(*directories)
+        func(path)
+
+    return inner
+
+
+@create_path
+def create_directory(path_to_implement: str) -> None:
+    try:
+        os.makedirs(path_to_implement)
```
Oleksl888, line 44: here you can call makedirs with exist_ok flag to handle case when directories exist

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+full_command = sys.argv
+command, *another_part_of_command = full_command[1:]
+
+if command == "-f":
+    create_file(another_part_of_command)
+
+elif command == "-d":
+    path_to_create = []
+
+    for directory in another_part_of_command:
```
Oleksl888, line 58: double check if your code handles invalid commands. e.g. `create_file.py -d dir1 dir2 -f`

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+                continue
+            if content == "stop":
+                break
+
+            creating_file.write(f"{line_num} {content}\n")
+            line_num += 1
+
+
+def create_path(func: Callable) -> Callable:
+    def inner(directories: list) -> None:
```
Oleksl888, line 34: it's nice that you made a decorator but calling one operation to join a directory does not really require a separate function

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+import datetime
+from typing import Callable
+
+
+def create_file(
+        name_of_the_file: str,
+        create_in_directory: list = None
+) -> None:
+
+    first_line = datetime.datetime.now()
```
Oleksl888, line 12: collecting input from user should be placed in a separate function

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+
+    if create_in_directory:
+        create_in_directory.append(name_of_the_file)
+        name_of_the_file = os.path.join(*create_in_directory)
+
+    with open(name_of_the_file, "w") as creating_file:
+        creating_file.write(first_line.strftime("%Y-%m-%d %H:%M:%S\n"))
+        line_num = 1
+        while True:
```
Oleksl888, line 21: I would suggest collecting data in an infinite loop outside an open file handle - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+            content = input("Enter content line: ")
+            if len(content) == 0:
+                print("Cannot add empty line")
```
Oleksl888, line 24: no need for this print

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+        create_in_directory: list = None
+) -> None:
+
+    first_line = datetime.datetime.now()
+
+    if create_in_directory:
+        create_in_directory.append(name_of_the_file)
+        name_of_the_file = os.path.join(*create_in_directory)
+
+    with open(name_of_the_file, "w") as creating_file:
```
Oleksl888, line 18: your file overwrites contents every time while it has to append to previous content. Also wrap file operations with try except to catch possible errors

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,75 @@
...
+        with open(name_of_the_file, "a") as creating_file:
+            creating_file.write(first_line.strftime("%Y-%m-%d %H:%M:%S\n"))
+    except Exception as e_info:
+        return e_info
+
+    line_num = 1
+    while True:
+        content = input("Enter content line: ")
+        if content == "stop":
+            break
```
oliasymak24, line 32: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
oliasymak24, line 1: Try to use argparse module

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #929, opened 2024-08-08
first_try

yashkunn:


app/create_file.py
```diff
@@ -1 +1,72 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+from typing import Any
+
+
+def create_directory(dirs: Any) -> None:
+    path = os.path.join(*dirs)
+    if not os.path.exists(path):
```
oliasymak24, line 9: u can use makedirs argument "exist_ok" here

app/create_file.py
```diff
@@ -1 +1,72 @@
...
+        mode = "a"
+    else:
+        mode = "w"
+
+    print("Enter content lines (type 'stop' on a new line to finish):")
+    lines = []
+    while True:
+        line = input("Enter content line: ")
+        if line.strip().lower() == "stop":
+            break
```
oliasymak24, line 27: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

oliasymak24 (changes requested):

LiudmylaKulinchenko (changes requested):
Try to use `argparse` module here — it would be easier, more readable, and you won't have to reinvent the wheel

app/create_file.py
```diff
@@ -1 +1,70 @@
...
+    path = os.path.join(*dirs)
+    os.makedirs(path, exist_ok=True)
+    print(f"Directory created or already exists: {path}")
+
+
+def create_file(file_path: Any) -> None:
+    if os.path.exists(file_path):
+        mode = "a"
+    else:
+        mode = "w"
```
oliasymak24, line 17: You can use only "a" mode

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #932, opened 2024-08-10
Solution Politov

Alexandr-Politov:


app/create_file.py
```diff
@@ -1 +1,45 @@
-# write your code here
+import datetime
+import os
+import sys
+
+
+def create_dir(directories: list) -> None:
+    path_to_file = os.path.join(*directories)
+    if not os.path.exists(path_to_file):
```
oliasymak24, line 8: u can use makedirs argument "exist_ok" here

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #936, opened 2024-08-11
solution (demo)

Narberal90:
This is not the best solution. I want to know in which direction I should move

### DIFF
--- a/app/create_file.py
+++ b/app/create_file.py
@@ -1,1 +1,37 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+list_args = sys.argv
+
+
+def write_file(path: str) -> None:
+    with open(path, "w") as file:
+        now = datetime.now().strftime("%Y-%m-%d  %H:%M:%S")
+        index = 1
+        file.write(now)
+        while True:
+            content = input("Enter content line: ")
+            if content == "stop":
+                break
+            file.write(f"\n{index} Line {content}")
+            index += 1
+
+
+if "-d" in list_args and "-f" in list_args:
+    file_index = list_args.index("-f")
+    dirs_list = list_args[list_args.index("-d") + 1:file_index]
+    dirs_path = os.path.join(*dirs_list)
+    os.makedirs(dirs_path, exist_ok=True)
+    file_name = list_args[file_index + 1]
+    file_path = os.path.join(dirs_path, file_name)
+    write_file(path=file_path)
+
+elif "-f" in list_args:
+    file_path = list_args[-1]
+    write_file(path=file_path)
+
+elif "-d" in list_args:
+    dirs_list = list_args[list_args.index("-d") + 1:]
+    dirs_path = os.path.join(*dirs_list)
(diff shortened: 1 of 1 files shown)


app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+from datetime import datetime
+import os
+import sys
+
+
+def write_file(path: str) -> None:
+    with open(path, "w") as file:
```
Oleksl888, line 7: Your file overwrites contents every time while it has to append to previous content. 
Also wrap file operations with try except to catch possible errors

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+        now = datetime.now().strftime("%Y-%m-%d  %H:%M:%S")
+        index = 1
+        file.write(f"{now}\n")
```
Oleksl888, line 10: I suggest to place collecting user input in a separate function to make your code more organized

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+        while True:
```
Oleksl888, line 11: I would suggest collecting data in an infinite loop outside an open file handle - while it is fine in this script, it can be an antipattern based on the scenario. While loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+
+def create_dirs(args_list: str) -> str:
+    file_index = args_list.index("-f")
+    dirs_list = args_list[args_list.index("-d") + 1:file_index]
+    dirs_path = os.path.join(*dirs_list)
+    os.makedirs(dirs_path, exist_ok=True)
+    file_name = args_list[file_index + 1]
+    file_path = os.path.join(dirs_path, file_name)
+    print(type(file_path))
```
Oleksl888, line 26: no need to print this in final version of your code

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+    return file_path
+
+
+if __name__ == "__main__":
+    args = sys.argv
+    if "-d" in args and "-f" in args:
```
Oleksl888, line 32: parsing args can be placed in a separate function. Plus double check if your code handles wrong input

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,56 @@
...
+def write_file(path: str, content_lines: list[str]) -> None:
+    try:
+        with open(path, "a") as file:
+            now = datetime.now().strftime("%Y-%m-%d  %H:%M:%S")
+            index = 1
+            file.write(f"{now}\n")
+            for content in content_lines:
+                file.write(f"{index} Line {content}\n")
+                index += 1
+            file.write("\n")
```
oliasymak24, line 25: You can try to use writelines instead

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #937, opened 2024-08-11
solution

HalynaRohatska:


app/create_file.py
```diff
@@ -1 +1,51 @@
-# write your code here
+import os.path
+import sys
+import datetime
+
+
+def create_directory(path: str) -> None:
+    os.makedirs(path, exist_ok=True)
```

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+def create_file(path_to_file: str) -> None:
+    with open(path_to_file, "a") as file:
+        timestamp = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{timestamp}\n")
+
+        line = 1
+        while True:
+            line_content = input("Enter content line: ")
+            if line_content.lower() == "stop":
+                break
```
oliasymak24, line 15: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py

oliasymak24: Try to use argparse module here

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #941, opened 2024-08-11
Solution

Bogdan-Rylov:


app/create_file.py

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+) -> str | None:
+    if dirs and filename:
+        return os.path.join(os.path.join(*dirs), filename)
+    elif filename:
+        return filename
+    return None
+
+
+def write_to_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
```
Oleksl888, line 37: Wrap operations with file in try except to catch all possible file-related errors

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        if os.stat(file_path).st_size != 0:
+            file.write("\n")
+
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+        while True:
+            new_line = input("Enter content line: ")
+            if new_line == "stop":
+                break
+            file.write(new_line + "\n")
```
Oleksl888, line 46: This can be optimized more:
1. Place collecting data from user in a separate function.
2.  Prepare data in a way that you don't have call `write` many times in a loop but write data all at once in one operation.
3. Opening an infinite loop inside of open file handle may be an antipattern is some cases - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+
+
+def write_to_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        if os.stat(file_path).st_size != 0:
+            file.write("\n")
+
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S\n"))
+        while True:
+            new_line = input("Enter content line: ")
```
Oleksl888, line 43: Output should be enumerated:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #942, opened 2024-08-11
Task Solution

achalaidiuk:


app/create_file.py
```diff
@@ -1 +1,42 @@
-# write your code here
+import argparse
+import os
+import datetime
+
+
+def create_directory(directory_name: str) -> None:
+    os.makedirs(directory_name, exist_ok=True)
```

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+
+    with open(file_name, "a") as file:
+        daytime = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{daytime}\n")
+        line = 1
+
+        while True:
+            text = input("Enter your text: ")
+            if text == "stop":
+                break
```
oliasymak24, line 19: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+import datetime
+
+
+def create_file(file_name: str, directory_name: str = None) -> None:
+    if directory_name:
+        os.makedirs(directory_name, exist_ok=True)
+        os.chdir(directory_name)
+
+    with open(file_name, "a") as file:
+        daytime = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
Oleksl888, line 12: I suggest to place collecting user input in a separate function. Also you need to revise how the lines are numbered - your output differs from one in the task (missing a number in the beginning of the line + you have extra blank spaces between lines). This is how it should like:
```
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

File: app/create_file.py [hunk 4]
```diff
@@ -1 +1,40 @@
...
+        file.write(f"{daytime}\n")
+        line = 1
+
+        while True:
```
Oleksl888, line 16: I suggest collecting data in an infinite loop outside an open file handle - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+            text = input("Enter your text: ")
+            if text == "stop":
+                break
+            file.write(f"Line{line} {text}\n")
```
Oleksl888, line 20: Calling `write` many times is not optimal - you can collect input in a list and then write it to file with one call with 'writelines`

Oleksl888 (changes requested):

app/create_file.py
achalaidiuk: About creating function, Olga said that : "No need to create a new function to use another one".

app/create_file.py
```diff
@@ -1 +1,45 @@
-# write your code here
+import argparse
+import os
+import datetime
+
+
+def write_text_to_file() -> list[str]:
```
Oleksl888, line 6: `collect user input` or similar would be a better name for this function.
Make sure this function gets all the information before writing to file. So adding timestamp, enumerating lines should happen here. Then you just call `writelines` in your `create_file` function once

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    if directory_name:
+        os.makedirs(directory_name, exist_ok=True)
+        os.chdir(directory_name)
+
+    with open(file_name, "a") as file:
+        daytime = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{daytime}\n")
+        lines = write_text_to_file()
+        file.writelines(lines)
+        file.write("\n")
```
Oleksl888, line 28: should be only one write here + you are opening an infinite loop inside of the open file. Collect data before you open the file for writing

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        line_num += 1
+    return lines
+
+
+def create_file(file_name: str, directory_name: str = None) -> None:
+    if directory_name:
+        os.makedirs(directory_name, exist_ok=True)
+        os.chdir(directory_name)
+
+    with open(file_name, "a") as file:
```
Oleksl888, line 23: You should wrap working with file in try-except to catch all possible file errors

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        file.write("\n")
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", "--directory", nargs="+")
+    parser.add_argument("-f", "--file")
+    args = parser.parse_args()
+
+    directory_name = os.path.join(*args.directory) if args.directory else None
```
Oleksl888, line 37: you don't need this ternary since argparser will throw an error if args will not be passed

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+        with open(file_name, "a") as file:
+            file.writelines(lines)
+    except Exception as e:
+        print(f"Error: {e}")
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", "--directory", nargs="+", required=True)
+    parser.add_argument("-f", "--file", required=True)
```
Oleksl888, line 37: -f is an optional argument - program should allow create directories if only -d flag is passed

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+    try:
+        with open(file_name, "a") as file:
+            file.writelines(lines)
+    except Exception as e:
+        print(f"Error: {e}")
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", "--directory", nargs="+", required=True)
```
Oleksl888, line 36: same here - if you pass only -f flag which is legal command your program will exit with error

app/create_file.py
```diff
@@ -1 +1,47 @@
...
+def create_file(file_name: str, directory_name: str) -> None:
+    if directory_name:
+        os.makedirs(directory_name, exist_ok=True)
+        os.chdir(directory_name)
+
+    lines = collect_user_input()
+    try:
+        with open(file_name, "a") as file:
+            file.writelines(lines)
+    except Exception as e:
```
Oleksl888, line 30: you better use OSError - it is more speciifc to work with filesystem

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #943, opened 2024-08-12
Solution

dpiuro:


app/create_file.py
```diff
@@ -1 +1,44 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory(path: str) -> None:
+    os.makedirs(path, exist_ok=True)
```

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,40 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def get_mode(file_path: str) -> str:
+    return "a" if os.path.exists(file_path) else "w"
```
oliasymak24, line 7: You can use only "a" mode

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+
+def write_content_to_file(file_path: str) -> None:
+    with open(file_path, get_mode(file_path)) as f:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        f.write(f"\n{timestamp}\n")
+        line_number = 1
+        while True:
+            content = input("Enter content line: ")
+            if content.lower() == "stop":
+                break
```
oliasymak24, line 18: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
oliasymak24, line 1: try to use argparse here

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #944, opened 2024-08-12
Solution

andreykolomiec:


pavlopro (changes requested):
How does your code works? It's just a defintions of functions

python create_file [-d directory_path] [-f file_name]

LLkaia (changes requested):
Program works well with that command
`python create_file.py -f file.txt -d dir1 dir2`
But is doing smth strange with this
`python create_file.py -d dir1 dir2 -f file.txt`

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+        )
+    else:
+        file_path = file_name
+
+    if file_name:
+        create_file(file_path)
+
+
+if __name__ == "__main__":
+    main()
```
Dimosphen1, line 65: - In the case the following command is executed, the file is not created
```
python create_file.py -f file.txt -d d1 d2 d3
```
- Make sure that console input is the same as in the task description, it should not enumerate lines during input;
- There should not be one blank line at the beginning of the newly created file;
- There should not be one blank line at the end of the newly created file.

Dimosphen1 (changes requested):

pavlopro (changes requested):
Doesn't work at all

pavlopro (changes requested):
There should be a blank line between different input sessions

PR #948, opened 2024-08-12
Solution

NazarSerdiuk1:


app/create_file.py
```diff
@@ -1 +1,46 @@
...
+            file_path = os.path.join(path, file_name)
+            create_file(file_path)
+        else:
+            dirs = args[d_index + 1:]
+            path = os.path.join(*dirs)
+            create_directories(path)
+    elif "-f" in args:
+        f_index = args.index("-f")
+        file_name = args[f_index + 1]
+        create_file(file_name)
```
LLkaia, line 42: GJ, but you need to handle case when `-f` will be passed before `-d`
You can take args values for `-d` until `-f` (like you did) or until end of list of arguments if `-f` appears before `-d`

LLkaia (changes requested):
Also consider case when key `-f or -d` provided but values for it not. In your implementation error will occur.

pavlopro (changes requested):
After input session, file has two blank lines at the end There shouldn't be any blank lines

With command `python create_file.py -d dir1 dir2` I receive message 
`Filename not provided. Use the -f option to specify the filename.`
It should work without messages when I just want to create a folder

pavlopro (changes requested):
Now it doesn't work at all

pavlopro (changes requested):
python create_file.py -f test.txt -d dir1 dir2

This gives me an error

pavlopro (changes requested):
Did you run your code ?
It produces an error when `python create_file.py -f test.txt -d dir1 dir2`

app/create_file.py
```diff
@@ -1 +1,50 @@
...
+
+def get_command() -> None:
+    if "-d" in command and "-f" not in command:
+        path_dest = os.path.join(*command[command.index("-d") + 1 :])
+        create_directory(path_dest)
+    elif "-d" in command and "-f" in command:
+        index_d = command.index("-d")
+        index_f = command.index("-f")
+
+        path_dest = os.path.join(*command[index_d + 1 : index_f])
```
LLkaia, line 16: command `python create_file.py -f file.txt -d dir1 dir2` still not working. The reason is because -f staying before -d.
the simliest way to fix it:
```
        if index_f > index_d:
            path_dest = os.path.join(*command[index_d + 1: index_f])
        else:
            path_dest = os.path.join(*command[index_d + 1:])
```

LLkaia (approved):
In general, your solution implemented well, but, please, pay attention to mentor's comments in the next time.

PR #949, opened 2024-08-12
task done

Sergi0bbb:


app/create_file.py
```diff
@@ -1 +1,61 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+def create_directory(directory_path: str) -> None:
+    os.makedirs(directory_path)
```

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+    with open(file_path, "a") as f:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        f.write(f"\n{timestamp}\n")
+
+        print("(to stop the program write 'stop')")
+        line_counter = 1
+        while True:
+            line = input("Enter content line: ")
+            if line.lower() == "stop":
+                break
```
oliasymak24, line 21: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,61 @@
...
+from datetime import datetime
+
+
+def create_directory(directory_path: str) -> None:
+    os.makedirs(directory_path)
+    print(f"{directory_path} directory created.")
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as f:
```
oliasymak24, line 12: f is a bad variable name

oliasymak24 (changes requested):

app/create_file.py
oliasymak24, line 1: Try to use argparse here

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
-# write your code here
+import argparse
+import os
+from datetime import datetime
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
```
Oleksl888, line 7: this block needs to be wrapped in try-except to catch file-related errors

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
```
Oleksl888, line 8: I suggest putting collection of data in separate function. Also instead of calling `write` multiple times, try to collect data and call `writelines` once for optimization

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        file.write(f"\n{timestamp}\n")
+
+        print("(to stop the program write 'stop')")
```
Oleksl888, line 11: remove unnecessary prints, keep only those that notify about an error

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        line_counter = 1
+        while True:
```
Oleksl888, line 13: Opening an infinite loop inside of open file handle may be an antipattern is some cases - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", "--dir", nargs="*", help="Directory path")
+    parser.add_argument("-f", "--file", help="File path")
+    args = parser.parse_args()
+
+    directory_path = ""
+
+    if args.dir:
+        directory_path = os.path.join(*args.dir)
+        os.makedirs(directory_path)
```
Oleksl888, line 34: Your code crashes when directory already exists

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+                file.write("\n")
+                break
+            file.write(f"{line_counter} {line}\n")
+            line_counter += 1
+
+    print(f"File {file_path} created successfully.")
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
```
Oleksl888, line 25: parsing args should be a separate function

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        create_file(file_path)
+
+    if not args.dir and not args.file:
+        parser.error("Neither '-d' nor '-f' flag provided.")
+
+
+if __name__ == "__main__":
+    try:
+        main()
+    except ValueError as e:
```
Oleksl888, line 48: why are you catching valuerror here?

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+    with open(file_path, "a") as file:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"\n{timestamp}\n")
+
+        print("(to stop the program write 'stop')")
+        line_counter = 1
+        while True:
+            line = input("Enter content line: ")
+            if line.lower() == "stop":
+                file.write("\n")
```
Oleksl888, line 16: your output file has a blank line in the beginning and there shouldn't be one

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+
+    directory_path = ""
+
+    if args.dir:
+        directory_path = os.path.join(*args.dir)
+        try:
+            os.makedirs(directory_path)
+            print(f"{directory_path} directory created.")
+        except FileExistsError:
+            print(f"{directory_path} directory already exists.")
```
oliasymak24, line 48: Use exist_ok for makedirs

oliasymak24 (changes requested):

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #950, opened 2024-08-12
Solution

mnezbrytska:


oliasymak24 (changes requested):

tests/test_main.py
LiudmylaKulinchenko, line 1: No need to delete this file 🤔

LiudmylaKulinchenko (changes requested):
Now let's 'wrap it all' using `argparse` method 😍 (I mean just use `argparse` here to not reinvent the wheel)

tests/test_main.py
mnezbrytska: I deleted becouse was this comment, mayby I didn't understand correct
del

oliasymak24 (approved):

PR #952, opened 2024-08-12
Solution

pashawarganov:


app/create_file.py

app/create_file.py
```diff
@@ -1 +1,23 @@
...
+    file_path = os.path.join(path, sys.argv[sys.argv.index("-f") + 1])
+else:
+    file_path = sys.argv[sys.argv.index("-f") + 1]
+
+with open(file_path, "a") as file:
+    file.write(str(datetime.datetime.now()) + "\n")
+    while True:
+        content = input("Enter content line: ")
+        if content.lower() == "stop":
+            break
```
oliasymak24, line 20: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,23 @@
...
+    directories = sys.argv[sys.argv.index("-d") + 1:sys.argv.index("-f")]
+    path = os.path.join(*directories)
+    os.makedirs(path)
+
+    file_path = os.path.join(path, sys.argv[sys.argv.index("-f") + 1])
+else:
+    file_path = sys.argv[sys.argv.index("-f") + 1]
+
+with open(file_path, "a") as file:
+    file.write(str(datetime.datetime.now()) + "\n")
```
oliasymak24, line 16: Date time format is not the same as required

oliasymak24 (changes requested):

app/create_file.py
oliasymak24, line 1: Try to use argparse module here

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+            if content.lower() == "stop":
+                file.write("\n")
+                break
+            file.write(str(line_number) + " " + content + "\n")
+            line_number += 1
+    print("<<<File created>>>")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
```
Oleksl888, line 36: You can define all the in `main` function and call it here

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+import argparse
+from datetime import datetime
+import os
+
+
+def create_path() -> str:
+    if args.directories:
+        path_dir = os.path.join(*args.directories)
+        try:
+            os.makedirs(path_dir)
```
Oleksl888, line 10: you can use `exist_ok` flag here instead of this try except - no need to reinvent the wheel :smiley:

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+            return os.path.join(path_dir, args.file)
+    return args.file
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        file.write(datetime.now().strftime("%Y-%m-%d %H:%M:%S") + "\n")
+
+        line_number = 1
+        while True:
```
Oleksl888, line 25: I suggest putting collection of data in separate function. Plus opening an infinite loop inside of open file handle may be an antipattern is some cases  - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+            content = input("Enter content line: ")
+            if content.lower() == "stop":
+                file.write("\n")
+                break
+            file.write(str(line_number) + " " + content + "\n")
```
Oleksl888, line 30: try to collect all data in a list and then write it to file with writelines method instead of calling write many times

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+            line_number += 1
+    print("<<<File created>>>")
```
Oleksl888, line 32: Please remove print statements - you can keep only those that signal about an exception

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        except FileExistsError:
+            print("<<<Directories already exists>>>")
+
+        if args.file:
+            return os.path.join(path_dir, args.file)
+    return args.file
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
```
Oleksl888, line 21: this block needs to be wrapped in try-excpet to catch file-related errors

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        "-d",
+        "--directories",
+        nargs="+",
+        help="List of directories"
+    )
+    parser.add_argument("-f", "--file", type=str, help="Filename")
+    args = parser.parse_args()
+    path = create_path(args.directories, args.file)
+
+    if args.file:
```
Oleksl888, line 54: no need for this `if` - if nothing is passed after -f flag argparser will exit with an error

app/create_file.py
```diff
@@ -1 +1,59 @@
-# write your code here
+from __future__ import annotations
+import argparse
+from datetime import datetime
+import os
+
+
+def create_path(directories: list | None, filename: str | None) -> str:
```
Oleksl888, line 7: We dont expect None here - these args will always be of correct types - argparser should take care of that

app/create_file.py
```diff
@@ -1 +1,59 @@
-# write your code here
+from __future__ import annotations
```
Oleksl888, line 1: why do you need this import?

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        content_list.append(str(line_number) + " " + content + "\n")
+        line_number += 1
+
+    return content_list
+
+
+def create_file(file_path: str) -> None:
+    content = make_content()
+    try:
+        user_file = open(file_path, "a")
```
Oleksl888, line 34: use context manager when working with file

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        user_file.writelines(content)
+    except Exception as e:
```
Oleksl888, line 36: OSError would be more specific in this case

Oleksl888 (changes requested):

app/create_file.py
pashawarganov: yes. I have python version 3.9, so I need this import for '|' in annotation. But in this case, if delete '| None' - I don`t need this anymore

oliasymak24 (approved):

PR #953, opened 2024-08-12
first

egor250330:


app/create_file.py
```diff
@@ -1 +1,59 @@
...
+import os
+
+
+def create_directory(directories: list) -> None:
+    path = os.path.join(*directories)
+    os.makedirs(path, exist_ok=True)
+
+
+def create_in_file(file_name: str) -> None:
+    with open(file_name, "w") as f:
```
oliasymak24, line 13: I think you can use only "a" mode

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        i += 1
+
+    return new_dict
+
+
+def main() -> None:
+    directories, file_name = parse_args(sys.argv[1:])
+
+    if file_name and directories:
+        create_directory(directories)
```
Oleksl888, line 49: directories path is not being created correctly

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+                f.write("\n")
+                break
+            line_numbers += 1
+            f.write(f"{line_numbers} {input_user}\n")
+
+
+def parse_args(args: list) -> dict:
+    new_dict = {"-d": [], "-f": None}
+    i = 0
+    while i < len(args):
```
Oleksl888, line 29: making a while loop inside of a while loop is not optimal. you can parse args using list slicing instead

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+def main() -> None:
+    directories, file_name = parse_args(sys.argv[1:])
+
+    if file_name and directories:
+        create_directory(directories)
+        full_name = os.path.join(*directories, file_name)
+        create_in_file(full_name)
+    elif file_name:
+        create_in_file(file_name)
+    elif directories:
```
Oleksl888, line 54: Can you try to simplify this to two `ifs`: 
1. if dir flag - create dir
2. 2. then if file flag is present - create file

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+
+
+def create_directory(directories: list) -> None:
+    path = os.path.join(*directories)
+    os.makedirs(path, exist_ok=True)
+
+
+def create_in_file(file_name: str) -> None:
+    with open(file_name, "a") as f:
+        f.write(datetime.now().strftime("%Y.%m.%d %H:%M:%S"))
```
Oleksl888, line 14: I suggest to place collecting user input in a separate function. 
Also you are calling `.write` many times - instead you could collect all data in a list and use `writelines` method.
Plus  I would suggest collecting data in an infinite loop outside an open file handle - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+import os
+
+
+def create_directory(directories: list) -> None:
+    path = os.path.join(*directories)
+    os.makedirs(path, exist_ok=True)
+
+
+def create_in_file(file_name: str) -> None:
+    with open(file_name, "a") as f:
```
Oleksl888, line 13: `f` is not a good file name + you should wrap work with file in try-excpet to catch all mistakes

app/create_file.py
```diff
@@ -1 +1,59 @@
...
+        f.write(datetime.now().strftime("%Y.%m.%d %H:%M:%S"))
+        line_numbers = 0
+
+        while True:
+            input_user = input("Enter content line: ")
+            if input_user == "stop":
+                f.write("\n")
+                break
+            line_numbers += 1
+            f.write(f"{line_numbers} {input_user}\n")
```
Oleksl888, line 23: i think lines are not numbered correctly at the moment

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,64 @@
...
+        lines.append(f"{line_numbers} {input_user}\n")
+
+    return lines
+
+
+def create_in_file(file_name: str) -> None:
+    lines = collect_user_input()
+
+    with open(file_name, "a") as file:
+        file.write(datetime.now().strftime("%Y.%m.%d %H:%M:%S\n"))
```
Oleksl888, line 30: you could actually include date and new line in your function `collect_user_input` and then you could call only method `writelines` instead of calling two extra writes

Oleksl888 (approved):
Well done 🎉

PR #955, opened 2024-08-13
implement py-create-file-from-terminal

Maksym-Turenko:


app/create_file.py
```diff
@@ -1 +1,62 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory(path: str) -> None:
+    os.makedirs(path, exist_ok=True)
```

app/create_file.py
oliasymak24, line 1: Try to use argparse here)

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,56 @@
-# write your code here
+import os
+from datetime import datetime
+import argparse
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a") as file:
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"\n{timestamp}\n")
```
oliasymak24, line 9: Your code add a new line at the beginning of the file, you don't need it

oliasymak24 (changes requested):

app/create_file.py
```diff
@@ -1 +1,63 @@
-# write your code here
+import os
+from datetime import datetime
+import argparse
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a+") as file:
+        file.seek(0)
```
Oleksl888, line 8: I suggest to place collecting user input in a separate function. Try to collect all data in a list and then pass it to create file function

app/create_file.py
```diff
@@ -1 +1,63 @@
...
+
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{timestamp}\n")
+
+        line_number = 1
+        while True:
+            line = input("Enter content line: ")
+            if line.strip().lower() == "stop":
+                break
+            file.write(f"{line_number} {line}\n")
```
Oleksl888, line 21: instead of calling `write` many times you can call `writelines` method once - but you need to be sure input is collected before that

app/create_file.py
```diff
@@ -1 +1,63 @@
...
+            line_number += 1
+
+    print(f"File '{file_path}' created/updated successfully.")
```
Oleksl888, line 24: remove debug prints from final version that you push to remote, you can keep only error messages as prints

app/create_file.py
```diff
@@ -1 +1,63 @@
...
+        file.seek(0)
+
+        if file.read(1):
+            file.write("\n")
+
+        timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        file.write(f"{timestamp}\n")
+
+        line_number = 1
+        while True:
```
Oleksl888, line 17: I suggest collecting data in an infinite loop outside an open file handle - while it is fine here, while loop could potentially run indefinitely waiting for user input and so the file could remain open for indefinite period leaving it unaccessable to other programs and consuming system resources

Oleksl888 (changes requested):

app/create_file.py
```diff
@@ -1 +1,63 @@
-# write your code here
+import os
+from datetime import datetime
+import argparse
+
+
+def create_file(file_path: str) -> None:
+    with open(file_path, "a+") as file:
```
Oleksl888, line 7: wrap the block of code where you work with file in try except to catch possible eerors

oliasymak24 (approved):

PR #957, opened 2024-08-20
Solution

zave52:


NazarSheremet (changes requested):
It would be better to split this code into functions: one that parses the data from the command line and another for creating and writing files.

YuriiSmolii (approved):
Well done

PR #962, opened 2024-08-26
Solution

Anon920:


app/create_file.py
```diff
@@ -1 +1,42 @@
...
+        for i, line in enumerate(content):
+            exist_file.write(f"{i + 1} {line} \n")
+        exist_file.write("\n")
+
+
+data_file = sys.argv[1:]
+print(data_file)
+
+if "-f" not in data_file:
+    make_directory(os.path.join(*data_file[1:]))
```
LLkaia, line 32: if both -f and -d missed, script will fail on this step

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+elif "-d" not in data_file:
+    create_file(data_file[-1])
+elif "-d" and "-f" in data_file:
+    if data_file.index("-d") < data_file.index("-f"):
+        file_dir = make_directory(os.path.join(*data_file[1:3]))
```
LLkaia, line 37: what if user will write more then 2 names of directories

LLkaia (changes requested):
Additionally let's wrap arguments parsing in separate function which will take args as parameter

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+            dir_path = make_directory(os.path.join(*data_file[3:]))
+            file_path = os.path.join(dir_path, data_file[1])
+        create_file(file_path)
+    elif "-f" in data_file:
+        create_file(data_file[-1])
+    elif "-d" in data_file:
+        make_directory(os.path.join(*data_file[1:]))
+    else:
+        print("Invalid arguments. Please provide"
+              " '-d' for directory or '-f' for file.")
```
LLkaia, line 45: This block of code is responsible for parsing argument so put it in separate function and call it with arguments

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+        for i, line in enumerate(content):
+            exist_file.write(f"{i + 1} {line} \n")
+        exist_file.write("\n")
+
+    data_file = sys.argv[1:]
+    print(data_file)
+
+    if "-f" in data_file and "-d" in data_file:
+        if data_file.index("-d") < data_file.index("-f"):
+            dir_path = make_directory(os.path.join(*data_file[1:3]))
```
LLkaia, line 33: instead of using hardcoded 1:3, use index of -f (it will be end of -d args)

LLkaia (changes requested):
Additionally try to run app before push to verify that every edge cases work correctly.

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+    elif "-f" in args:
+        create_file(args[-1])
+    elif "-d" in args:
+        make_directory(os.path.join(*args[args.index("-d") + 1:]))
+    else:
+        print("Invalid arguments. Please provide"
+              " '-d' for directory or '-f' for file.")
+
+
+def main(*args) -> Any:
```
vasylhnatiuk, line 51:
```suggestion
def main(*args) -> None:
```

vasylhnatiuk: the same everywhere if code returns nothing

app/create_file.py
```diff
@@ -1 +1,54 @@
-# write your code here
+import sys
+import os
+import datetime
+from typing import Any
+
+
+def make_directory(name_dir: str) -> Any:
```
vasylhnatiuk, line 7: Any is wrong annotation! use more specific type of data

vasylhnatiuk (changes requested):

PR #963, opened 2024-08-30
solution

nrjss67:


vsmutok (changes requested):
image

app/create_file.py
```diff
@@ -1 +1,44 @@
...
+            if ("." not in part
+                    and "-" not in part):
+                part_of_dir.append(part)
+            if "." in part:
+                file_name = part
+
+        joined_path = os.path.join(*part_of_dir)
+        os.makedirs(joined_path, exist_ok=True)
+
+        content_writer(joined_path + "/" + file_name)
```
vsmutok, line 32: better to use `os.path.join`

vsmutok (changes requested):
image
there must be one blank line between blocks of text

PR #968, opened 2024-09-03
Fixed error when only "-d" in command

arsen-arutiunov:


vasylhnatiuk (changes requested):
image
there should be empty line between 3 and 4 lines of data

PR #974, opened 2024-09-04
solution

mlupekha:


YukihiroSM (changes requested):
image
If the file option goes before directories - directories are not created and file is created in place of command execution. Fix it, please.

mlupekha:
added 1 + 1 == 2, to request another review

vasylhnatiuk (approved):

PR #981, opened 2024-09-06
Fixed error when in command only "-d"

arsen-arutiunov:


vasylhnatiuk (changes requested):
image
how did you test your code? it does nothing

PR #983, opened 2024-09-06
Solution

arsen-arutiunov:


arsen-arutiunov:
Why is it that when I check it works and you don't?

arsen-arutiunov:
I've now checked all the commands again, they work. Could it not work because I have a problem with the interpreter, I check by typing each command in the code, so can't run the command itself through the terminal. (after re-installing the project from Github it does not install correctly and I get errors).

PR #986, opened 2024-09-12
Solution

juliazelena:


app/create_file.py
```diff
@@ -1 +1,27 @@
-# write your code here
+import sys
+import os
+import datetime
+
+
+for arg in sys.argv:
+    if arg == "-d":
+        if "-f" not in sys.argv:
+            target_path = os.path.join("/".join(sys.argv[2:]))
```
LLkaia, line 9: use index of -d instead of hardcoding index

app/create_file.py
```diff
@@ -1 +1,27 @@
...
+        else:
+            target_path = os.path.join("/".join(sys.argv[2:-2]))
+        if not os.path.exists(target_path):
+            os.makedirs(target_path, exist_ok=True)
+        os.chdir(target_path)
+
+filename = sys.argv[-1] if "-f" in sys.argv else "file.txt"
```
LLkaia, line 16: What -f will stand before -d?

LLkaia (changes requested):

pavlopro (changes requested):
python create_file.py -d dir1 dir2 -f test.txt

1. It creates me file with name `file.txt`. But should create `test.txt`
2. There shouldn't be any blank lines at the end of an created file

pavlopro (changes requested):
There shouldn't be any blank lines at the end of created file during input session

LLkaia (approved):
LGTM

PR #997, opened 2024-09-16
py-create-file-from-terminal

s13en:
Solution


LLkaia (changes requested):
Stop is not working
image

LLkaia (approved):
Your program looks awesome!

PR #1002, opened 2024-09-20
Solution

Jejemira:


app/create_file.py
```diff
@@ -1 +1,65 @@
...
+        else:
+            time_output = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+            source_file.write(f"{time_output}\n")
+            page_number = 1
+            while True:
+                line_content = input("Enter content line: ")
+                if line_content.lower() == "stop":
+                    break
+                source_file.write(f"{page_number} {line_content}\n")
+                page_number += 1
```
LLkaia, line 29: Try to avoid repetition of same code

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+
+
+def create_directory() -> None:
+    if "-f" in argv and "-d" in argv and argv.index("-f") > argv.index("-d"):
```
LLkaia, line 33: since you are reusing -d and -f indexes few times, create separate variables for them

app/create_file.py
```diff
@@ -1 +1,65 @@
...
+            page_number = 1
+            while True:
+                line_content = input("Enter content line: ")
+                if line_content.lower() == "stop":
+                    break
+                source_file.write(f"{page_number} {line_content}\n")
+                page_number += 1
+
+
+def create_directory() -> None:
```
LLkaia, line 32: This is wrong name of function since it is not only creating directory. Create function main, which will parse args and will call function to create directory or to create file

LLkaia (changes requested):
We should not see blank line at the start of the file:
image

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+    with open(file_name, "a") as source_file:
+        time_output = datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        source_file.write(f"{time_output}\n")
+        page_number = 1
+        while True:
+            line_content = input("Enter content line: ")
+            if line_content.lower() == "stop":
+                break
+            source_file.write(f"{page_number} {line_content}\n")
+            page_number += 1
```
LLkaia, line 20: Better, but still open file twice looks bad.

For your approach, create variable `file_exists = os.path.exists(file_name)` and then use when file is opened to write first line.

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+
+
+def main() -> None:
+    char_f = "-f"
+    char_d = "-d"
```
LLkaia, line 25:
```suggestion
    index_f = argv.index("-f")
    index_d = argv.index("-d")
```

And then reuse it

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+        directory_names = os.path.join(*argv[2:argv.index(char_f):])
+        file_name = os.path.join(*argv[argv.index(char_f) + 1::])
+        if os.path.isdir(directory_names):
+            os.chdir(directory_names)
+            create_file(file_name)
+
+        else:
+            os.makedirs(directory_names)
+            os.chdir(directory_names)
+            create_file(file_name)
```
LLkaia, line 37: you can use `exist_ok=True` for os.makedirs to not rise error if dir exists.

app/create_file.py
```diff
@@ -1 +1,62 @@
...
+        directory_names = os.path.join(*argv[argv.index(char_d) + 1::])
+        file_name = os.path.join(*argv[2:argv.index(char_d):])
+        if os.path.isdir(directory_names):
+            os.chdir(directory_names)
+            create_file(file_name)
+
+        else:
+            os.makedirs(directory_names)
+            os.chdir(directory_names)
+            create_file(file_name)
```
LLkaia, line 50: same here

LLkaia (changes requested):

vasylhnatiuk (approved):

PR #1006, opened 2024-09-26
Solution

Viacheslav417865:


Viacheslav417865:
py-create-file-from-terminal 1

Viacheslav417865:
py-create-file-from-terminal 2

PR #1011, opened 2024-10-04
solution

dmytrik:


app/create_file.py
```diff
@@ -1 +1,51 @@
...
+        new_file.write(
+            str(datetime.now().strftime("%Y-%m-%d %H:%M:%S")) + "\n"
+        )
+        count = 1
+        while True:
+            line = input("Enter content line: ")
+            if line == "stop":
+                break
+            new_file.write(f"{str(count)} " + line + "\n")
+            count += 1
```
AnyoneClown, line 43: Simplify it. You can check if file exists with `if os.path.exists(path):` for this. Don't use global variable

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+        try:
+            os.makedirs(path)
+        except Exception:
+            pass
+        current_path = os.path.join(*way)
+        create_file(current_path)
+        return
+
+    path = os.path.join(*way)
+    os.makedirs(path)
```
AnyoneClown, line 22: You are calling this function only if you have flag `-d`. Why do you expect `-f` to be here? Refactor it

app/create_file.py
```diff
@@ -1 +1,51 @@
...
+            new_file.write(f"{str(count)} " + line + "\n")
+            count += 1
+
+
+current_directory, flag, *expected_way = sys.argv
+
+if flag == "-d":
+    create_directory(expected_way)
+if flag == "-f":
+    create_file(*expected_way)
```
AnyoneClown, line 51: Create `main` function for you logic and refactor it using:

```
if __name__ == "__main__":
    main()
```

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM!

PR #1015, opened 2024-10-07
Solution

Atikiho:


dmytrik (approved):

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+path = os.path.join(*(command[2:]))
+path_for_2_flags = os.path.join(*(command[2:-2]))
+
+
+def create_dirs() -> None:
+    if not os.path.exists(path_for_2_flags):
+        if "-f" in command:
+            os.makedirs(path_for_2_flags)
+        else:
+            os.makedirs(path)
```
AnyoneClown, line 15: You are calling this function only when your flag is `-d`. So, how do you expect flag `-f` to be here?

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+
+
+def create_file() -> None:
+    if "-d" in command:
+        os.chdir(path_for_2_flags)
```
AnyoneClown, line 20: The same here

app/create_file.py
```diff
@@ -1 +1,37 @@
...
+    with open(command[-1], "a") as new_file:
+        new_file.write(datetime.datetime.now()
+                       .strftime("%Y-%m-%d %H:%M:%S") + "\n")
+        text = input("Enter content line:")
+        index = 0
+        while text != "stop":
+            index += 1
+            new_file.write(f"{index} {text}\n")
+            text = input("Enter content line:")
```
AnyoneClown, line 29: You don't need to create 2 `input` here. Just create 1 on start of your `while` loop

AnyoneClown (changes requested):

app/create_file.py
Atikiho: But there is cases when 2 flags are given and i need to change function behaviour in these cases

> If both -d and -f flags passed, app creates directory and file with content inside this directory.

AnyoneClown: Yeah, my bad. Missed this point :)

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+                break
+            new_file.write(f"{index} {text}\n")
+        new_file.write("\n")
+
+
+if "-d" in command:
+    create_dirs()
+
+if "-f" in command:
+    create_file()
```
AnyoneClown, line 38: My bad, but forgot to add this comment too...

Let's create entrypoint `main` func to do all this stuff.

Afther that we will call it inside ` if "__name__" == "__main__"`

AnyoneClown (approved):
Good, but fix comment above

PR #1016, opened 2024-10-08
created program for create file in the specified directory

Mxbely:


app/create_file.py
```diff
@@ -1 +1,48 @@
...
+                                                    "%H:%M:%S") + "\n")
+    with open("file_name", "a") as file:
+        while True:
+            inp = input("Enter content line: ")
+            if inp == "stop":
+                break
+            file.write(inp + "\n")
+
+
+input_arguments = sys.argv
```
AnyoneClown, line 38: sys.argv will work fine, but we can improve this program using `argparse`

[Link](https://docs.python.org/3/library/argparse.html)

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+input_arguments.pop(0)
+os.chdir("app")
+if "-d" not in input_arguments:
+    write_information_into_file()
+else:
+    directory = []
+    file_name = []
+    create_path_and_file_name(input_arguments)
+    create_dir()
+    write_information_into_file()
```
AnyoneClown, line 48: Create entrypoint function `main` for this. And then call it inside `if "__name__" == "__main__"`

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+
+
+input_arguments = sys.argv
+input_arguments.pop(0)
+os.chdir("app")
+if "-d" not in input_arguments:
+    write_information_into_file()
+else:
+    directory = []
+    file_name = []
```
AnyoneClown, line 45: You don't need to create this variables here. It's not recommended to use global variables in your code.

You can make return in your `func` to create list of directories

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+def write_information_into_file() -> None:
+    with open("file_name", "w") as file:
+        file.write(datetime.datetime.now().strftime("%Y-%m-%d "
+                                                    "%H:%M:%S") + "\n")
+    with open("file_name", "a") as file:
+        while True:
+            inp = input("Enter content line: ")
+            if inp == "stop":
+                break
+            file.write(inp + "\n")
```
AnyoneClown, line 35: If you are not checking if file exists already, why do u use 2 context managers here? We can use `while` loop inside first context manager

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        os.chdir(direc)
+
+
+def write_information_into_file() -> None:
+    with open("file_name", "w") as file:
+        file.write(datetime.datetime.now().strftime("%Y-%m-%d "
+                                                    "%H:%M:%S") + "\n")
+    with open("file_name", "a") as file:
+        while True:
+            inp = input("Enter content line: ")
```
AnyoneClown, line 32: Use propper naming. For example `content_line`

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+    flag = True
+    for value in arguments:
+        if value == "-f":
+            flag = False
+        if flag:
+            directory.append(value)
+        else:
+            file_name.append(value)
+    directory.pop(0)
+    file_name.pop(0)
```
AnyoneClown, line 16: If you are creating path and file_name here. We can simply return directories inside of a list + file_name here. Refactor without using `pop()` and global list

AnyoneClown (changes requested):

app/create_file.py
Mxbely: i spent a lot of time and couldn't understend, how use it correct, sorry 😩 
i tried to use smth like this (my example) and can't made a correct input 
and i decided use sys.argv (but i know, argparse also use sys.argv inside yourself)
```
    parser = argparse.ArgumentParser(description="create new file")
    parser.add_argument("-d", default="abc", type=str, help="-d for create directory, -f for create file")
    parser.add_argument("-file_name", default="file_name", type=str, help="-d for create directory, -f for create file")
    args = parser.parse_args()
    print(args.__dict__)
```

app/create_file.py
```diff
@@ -1 +1,48 @@
...
+        with open(file_name, "w") as file:
+            file.write(datetime.datetime.now().strftime("%Y-%m-%d "
+                                                        "%H:%M:%S") + "\n")
+            while True:
+                content_line = input("Enter content line: ")
+                if content_line == "stop":
+                    break
+                file.write(content_line + "\n")
+
+    input_arguments = sys.argv
```
Nattalli, line 36: example:

```python
parser = argparse.ArgumentParser()
parser.add_argument("-f", "--file", help="-d for create directory, -f for create file")
parser.add_argument("-d", "--dir", nargs="+", help="-d for create directory, -f for create file")
command = parser.parse_args()
```

Nattalli: and after r=that you can use it:
```
if command.dir:
     path = os.path.join(*command.dir)
     ...

if command.file:
     ...
```

Nattalli (changes requested):
I reject your solution because I honestly believe that you can do that with suggested statements. Also, you are ahead of the group so you have time for finding solution! 🌹

Nattalli (approved):
now it's prettier 💅🏻

PR #1017, opened 2024-10-12
solution

KKvladys:


app/create_file.py
```diff
@@ -1 +1,49 @@
...
+        index_f = commands.index("-f")
+        file_name = commands[index_f + 1:]
+
+        if "-d" in commands:
+            index_d = commands.index("-d")
+            name_dir = commands[index_d + 1:index_f]
+            dir_path = os.path.join(*name_dir)
+
+            if not os.path.exists(dir_path):
+                os.makedirs(dir_path)
```
vladyslav-tmf, line 18: This `if` check can be replaced by adding the `exist_ok=True` parameter, which prevents an error if the directory already exists. This is a more Pythonic approach.
```suggestion
            os.makedirs(dir_path, exist_ok=True)
```

vladyslav-tmf (changes requested):
No handling for the situation when no flags are entered will result in the program doing nothing and not giving the user any hints, which can be confusing and inconvenient. To improve the user experience, you should add a check and error message in case of missing arguments.

PR #1024, opened 2024-10-20
Solution

VladyslavBon:


skyfoxwork (approved):
I think good job!

I have some little tips.

When I create a file I get an extra blank line at the beginning of the file. I don't now it is a big problem or not.

and when I use:

    python3 create_file.py -f file.txt -d dir1 dir2

you program don't work. In description for task it is look right:

But I think it will better if it will work when you can change flags.

and you program crash when I use:

     python3 create_file.py

without flags or other arguments. It will better if program does not raise exception because it is like app.

PR #1028, opened 2024-10-21
solution

spa7id:


app/create_file.py
```diff
@@ -1 +1,50 @@
...
+
+    for i in range(1, len(sys.argv)):
+        if sys.argv[i] == "-d":
+            d_flag = True
+            directory = os.path.join(*sys.argv[i + 1:])
+            break
+        elif sys.argv[i] == "-f":
+            f_flag = True
+            filename = sys.argv[i + 1]
+            break
```
vladyslav-tmf, line 42: Your implementation uses a `break` statement in the loop when processing either the `-d` or `-f` flags, which prevents both flags from being handled simultaneously. To meet the task requirements, you should allow the loop to continue checking for both flags before proceeding with directory and file creation.

vladyslav-tmf (changes requested):

PR #1040, opened 2024-10-22
solution (?)

dxrrkwm:


pohgen (approved):
you need to add one extra "\n"
before the second timestamp

f.e.
2022-02-01 14:41:10
1 Line1 content
2 Line2 content
3 Line3 content

2022-02-01 14:46:01
1 Another line1 content
2 Another line2 content
3 Another line3 content

but overall solution is perfect

PR #1041, opened 2024-10-22
Soltuion

zaietsmo:


app/create_file.py
```diff
@@ -1 +1,63 @@
...
+    with open(file_path, "a") as f:
+        f.write(f"\n{timestamp}\n")
+
+        count = 1
+        while True:
+            content = input(f"Enter content line {count}: ")
+            if content.lower() == "stop":
+                break
+            f.write(f"{count} {content}\n")
+            count += 1
```
dxrrkwm, line 32: you can use these two functions to call it if name is main avoiding adding third (main) function

app/create_file.py
```diff
@@ -1 +1,63 @@
...
+        print(f"File '{file_path}' exists. Appending new content...")
+    else:
+        print(f"Creating new file: '{file_path}'")
+
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+
+    with open(file_path, "a") as f:
+        f.write(f"\n{timestamp}\n")
+
+        count = 1
```
dxrrkwm, line 26: it's better to use something like line number here

dxrrkwm (changes requested):

PR #1051, opened 2024-10-23
Solution

Bondzik-S:


FursykIvan (approved):
Looks good form me)

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+
+
+def get_file_content() -> None:
+    content_lines = []
+    while True:
+        line = input("Enter content line: ")
+        if line.lower() == "stop":
+            break
+        content_lines.append(line)
+    return content_lines
```
AnyoneClown, line 34: Are you sure, that you are returning `None` here?

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+    full_path = os.path.join(dir_path, filename)
+
+    with open(full_path, "a") as f:
+        if os.path.getsize(full_path) > 0:
+            f.write("\n")
+        f.write(f"{timestamp}\n")
+        f.write("\n".join(lines_with_numbers) + "\n")
+
+    if not is_directory and not is_file:
+        print("Please provide at least one flag (-d or -f).")
```
AnyoneClown, line 55: Create `main()` func. It will be an entrypoint for your programm. Then call it from `if __name__ == "__main__":` block

AnyoneClown (changes requested):
Your solution is good, but you might try to implement it with [argparse](https://docs.python.org/3/library/argparse.html). Also fix comments below

AnyoneClown (approved):
Good, but don't forget to delete `#` before pushing your code :)

PR #1113, opened 2024-12-22
Solution

julia4406:


taraskhom (changes requested):
When creating only directories, an error appears with the following message: "error: the following arguments are required: -f/--filename"

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+def get_command_from_terminal() -> tuple:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-d", "--directories", nargs="*",
+                        help="creates directories -d dir1 dir2")
+    parser.add_argument("-f", "--filename", default="file.txt", required=False,
+                        help="creates file -f filename")
+    args = parser.parse_args()
+    directories = args.directories
+    filename = args.filename
+    return directories, filename
```
Nattalli, line 15: not needed variables creation

```suggestion
    return args.directories, args.filename
```

app/create_file.py
```diff
@@ -1 +1,49 @@
...
+def file_manager() -> str:
+    dir_names, file_name = get_command_from_terminal()
+    dir_path = os.getcwd()
+    if dir_names:
+        for directory in dir_names:
+            dir_path = os.path.join(dir_path, directory)
+            if not os.path.exists(dir_path):
+                os.mkdir(dir_path)
+    file_path = os.path.join(dir_path, file_name)
+    return file_path
```
Nattalli, line 27:
```suggestion
    return os.path.join(dir_path, file_name)
```

Nattalli (changes requested):

L1nk3rrr (approved):
Well done!

PR #1115, opened 2024-12-22
Create task create file from terminal

ihorhalyskiy:


app/create_file.py
```diff
@@ -1 +1,68 @@
...
+    return os.path.join(root_path, *dirs_names)
+
+
+def generate_file(
+        path: str,
+        file_name: str,
+        content: str
+) -> None:
+
+    with open(os.path.join(path, file_name), "w", encoding="utf-8") as file:
```
ruslanyemets, line 49: You should change 'w' mode to another because you need to add content under existing content if file already exists.

ruslanyemets (approved):
Good job! But you need to make minor changes to meet the requirements of the task.

app/create_file.py
ihorhalyskiy: > You should change 'w' mode to another because you need to add content under existing content if file already exists.

thanks, i change )

PR #1132, opened 2024-12-23
solution_v.1.0

yelizaveta-zh:


bodiakof (approved):
Nicely done! Although, since you only need `argv` from `sys` might be better to import just it, not the entire module. Also, to improve readability, might be a good idea to divide `create_file()` into several functions

PR #1134, opened 2024-12-24
solution

MaksVakulenko:


julia4406 (approved):
Possibly, there is a logic mistake during file creating: command python app/create_file.py -d dir1 -f f1.txt creates folders dir1/-f/f1.txt/ and then creates file with name 'f1.txt'. Just check this.

PR #1152, opened 2025-01-11
Created function create_file

shafran-andriy:


pavlopro (changes requested):
The file shouldn't contain just

1 Line1 content
2 Line2 content
3 Line3 content
...

It should containt user's input

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+        file.write(f"{time_string}\n")
+    line = ""
+    count = 1
+    while line != "stop":
+        line = input(f"Enter content line: Line{count} content")
+        if line == "stop":
+            break
+        with open(f"{args.f[0]}", "a+") as file:
+            file.write(f"{count} {line}\n")
+        count += 1
```
vsmutok, line 27: Try to use `with` only once and open the file just once.  
It's also important that the data should be appended to the file, not overwritten.

app/create_file.py
```diff
@@ -1 +1,42 @@
...
+
+def create_file() -> None:
+    with open(f"{args.f[0]}", "w") as file:
+        pass
+    with open(f"{args.f[0]}", "a+") as file:
+        file.write(f"{time_string}\n")
+    line = ""
+    count = 1
+    while line != "stop":
+        line = input(f"Enter content line: Line{count} content")
```
vsmutok, line 22:
```suggestion
        line = input(f"Enter content line: Line{count} content: ")
```

vsmutok (changes requested):

app/create_file.py
shafran-andriy: fixed

app/create_file.py
```diff
@@ -1 +1,40 @@
...
+        file.write(f"{time_string}\n")
+    line = ""
+    count = 1
+    while line != "stop":
+        line = input(f"Enter content line: Line{count} content:")
+        if line == "stop":
+            break
+        with open(f"{args.f[0]}", "a+") as file:
+            file.write(f"{count} {line}\n")
+        count += 1
```
vsmutok, line 25: Try to use `with` only once and open the file just once.

vsmutok (changes requested):

app/create_file.py
shafran-andriy: rewrited

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+def create_file() -> None:
+    with open(f"{args.f[0]}", "a+") as file:
+        file.write(f"{time_string}\n")
+        count = 1
+        while True:
+            line = input(f"Enter content line: Line{count} content: ")
+            if line == "stop":
+                break
+            file.write(f"{count} {line}\n")
+            count += 1
```
vsmutok, line 23: There should be a blank line between entries in the file.
image

app/create_file.py
```diff
@@ -1 +1,38 @@
...
+
+if args.f is None:
+    create_dirs()
+else:
+    if args.d is None:
+        create_file()
+    else:
+        create_dirs()
+        os.chdir(os.path.join(*args.d))
+        create_file()
```
vsmutok, line 38: This will look somewhat better. :)
```suggestion
if args.f is None or args.d is not None:
    create_dirs()

if args.d is not None:
    os.chdir(os.path.join(*args.d))

if args.f is not None:
    create_file()
```

vsmutok (changes requested):

app/create_file.py
shafran-andriy: please explain why your code will be better

vsmutok: 1. **Readability**: Independent blocks are easier to understand than nested structures.  
2. **Avoids nesting**: No deep indentation, making the code cleaner.

app/create_file.py
```diff
@@ -1 +1,39 @@
...
```
vsmutok, line 39:
```suggestion
if args.f is None or args.d is not None:
    create_dirs()
    
if args.d is not None:
    os.chdir(os.path.join(*args.d))

if args.f is not None:
    create_file()
```

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1153, opened 2025-01-11
Solution

MarianaAB19:


vsmutok (changes requested):
When I run this command in the terminal, nothing happens.
```
```

vsmutok (changes requested):
1. The logic doesn't work as expected: an extra folder `-d` is created, and the file name is incorrect.  
2. If I run the command a second time, it causes an error.  

Tip: Read about the `argparse` module; it might be helpful for your task here.
image

vasylhnatiuk (approved):

PR #1197, opened 2025-02-12
Solution

ip-ilya:


SerhiyTo (changes requested):
Your solution looks good, but try to divide the function into several tribal ones, thereby simplifying the logic of your program. You can also use the `ArgumentParser` class from the `argparse` module to create more understandable commands instead of using the usual `if` statements.

PR #1204, opened 2025-02-14
Solution

kram3ko:


app/create_file.py
```diff
@@ -1 +1,46 @@
-# write your code here
+from __future__ import annotations
+
+import sys
+from datetime import datetime
+from pathlib import Path
+
+
+def makes_dir(path_file: Path) -> None:
+    path_file.mkdir(parents=True, exist_ok=True)
```
oliasymak24, line 9: No need to create a new function to use another one

app/create_file.py
```diff
@@ -1 +1,46 @@
...
+def write_data(file_name: Path) -> None:
+    current_time = datetime.now()
+    formatted_time = current_time.strftime("%Y-%m-%d %H:%M:%S")
+
+    with open(file_name, "a", encoding="utf-8") as file:
+        file.write(formatted_time + "\n")
+        while True:
+            line = input("Enter content line: ")
+            if line == "stop":
+                break
```
oliasymak24, line 21: add blank line if command == "stop". In the examples of the task when inserting new text, first comes the empty line, and then the new values

app/create_file.py
```diff
@@ -1 +1,46 @@
-# write your code here
+from __future__ import annotations
```
oliasymak24, line 1: you don't need this import

oliasymak24 (changes requested):

PR #1218, opened 2025-02-28
Dev

Sotnikov-100:


Nikita-pyth (approved):
I like a lot how you break down it into smaller functions

PR #1240, opened 2025-03-10
task solution

yarms1:


Nikita-pyth (approved):
It's great, but next time add *.txt to gitignore so the results of the program won't be added to commits✌️

PR #1254, opened 2025-03-13
solution

DHlomozda:


app/create_file.py
```diff
@@ -1 +1,51 @@
...
+def create_directory(dirs: list) -> None:
+    path = os.path.join(*dirs)
+    os.makedirs(path)
+
+
+def create_file(filename: str) -> None:
+    if os.path.exists(filename):
+        append_write = "a"
+    else:
+        append_write = "w"
```
Boyok1337, line 15: Also, you can define it in one line, cause it`s simple condition

Boyok1337 (approved):
Good one!

PR #1276, opened 2025-03-27
'Solution'

Sezaro104:


app/create_file.py
```diff
@@ -1 +1,54 @@
-# write your code here
+import sys
+import os
+from datetime import datetime
+
+
+def create_directory(path_parts: any) -> str:
```
Boyok1337, line 6: Can `path_parts` really be Any?
What if i pass dict?

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+    path = os.path.join(*path_parts)
+    os.makedirs(path, exist_ok=True)
+    return path
+
+
+def write_to_file(file_path: any) -> None:
```
Boyok1337, line 12: The same, maybe better to define `file_path` as `str`?

app/create_file.py
```diff
@@ -1 +1,54 @@
...
+        arg = args.pop(0)
+        if arg == "-d":
+            while args and args[0] != "-f":
+                dir_path.append(args.pop(0))
+        elif arg == "-f" and args:
+            file_name = args.pop(0)
+
+    if not file_name:
+        print("File name is required.")
+        return
```
Boyok1337, line 46: Try to simplify this logic.
Hard to read

Boyok1337 (changes requested):

PR #1303, opened 2025-04-20
v1

KrystianWojtiuk:


KrystianWojtiuk:
@mateacademy-ai-mentor creating file in current directory when -d is not provided is intended behaviour.

PR #1364, opened 2025-06-28
Solution

PavloReutskiy:


app/create_file.py
```diff
@@ -1 +1,56 @@
...
+
+
+def get_flag_index(args: list[str], flag: str) -> int | None:
+    if flag in args:
+        return args.index(flag)
+    return None
+
+
+d_flag_index = get_flag_index(args, "-d")
+f_flag_index = get_flag_index(args, "-f")
```
MNDonut, line 16: let's move these variables after the all functions so that it's easier to find them

PR #1372, opened 2025-07-04
Solution

zqpiq:


app/create_file.py
```diff
@@ -1 +1,52 @@
...
+    f_idx = sys.argv.index("-f")
+    dirs = sys.argv[d_idx + 1:f_idx]
+    filename = sys.argv[f_idx + 1]
+
+    if len(dirs) == 1:
+        os.mkdir(dirs[0])
+        str_dirs = dirs[0]
+    else:
+        str_dirs = os.path.join(*dirs)
+        os.makedirs(str_dirs, exist_ok=True)
```
L1nk3rrr, line 40: the same as `d_command` method

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+
+    file_path = os.path.join(str_dirs, filename)
+    with open(file_path, "w") as file:
+        writing(file)
```
L1nk3rrr, line 44: the same as `f_command`

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+
+elif "-f" in sys.argv:
+    f_idx = sys.argv.index("-f")
+    f_command(sys.argv[f_idx + 1])
+
+elif "-d" in sys.argv:
+    d_idx = sys.argv.index("-d")
+    d_command(sys.argv[d_idx + 1:])
```
L1nk3rrr, line 52: also this all bock can be moved to a separate method and must be called in construction like
```
if __name__ == "__main__":
    # your code here
```

It will be the best practice, as you want to run that code, only if it's the main file

app/create_file.py
```diff
@@ -1 +1,52 @@
...
+        str_dirs = os.path.join(*dirs)
+        os.makedirs(str_dirs, exist_ok=True)
+
+    file_path = os.path.join(str_dirs, filename)
+    with open(file_path, "w") as file:
+        writing(file)
+
+elif "-f" in sys.argv:
+    f_idx = sys.argv.index("-f")
+    f_command(sys.argv[f_idx + 1])
```
L1nk3rrr, line 48: can be symplified, you don't need f_idx variable, simply
`f_command(sys.argv[sys.argv.index("-f") + 1])`

app/create_file.py
L1nk3rrr, line 52: same about d_idx

L1nk3rrr (changes requested):
Please review my comments to improve your code!

app/create_file.py
```diff
@@ -46,7 +46,9 @@ def main() -> None:
         f_command(sys.argv[sys.argv.index("-f") + 1])
 
     elif "-d" in sys.argv:
-        d_command(sys.argv[sys.argv.index("-d") + 1])
+        d_idx = sys.argv.index("-d")
+        dirs = sys.argv[d_idx + 1:]
+        d_command(dirs)
```
MNDonut, line 51: what if the command is
`command -d dir1 dir2 dir3 -f file.txt`?
Your slice will take all words and flags

MNDonut (changes requested):

PR #1375, opened 2025-07-07
Solution

DrGregoryHouse4676:


app/create_file.py
```diff
@@ -1 +1,91 @@
...
+                directory_path.append(args[i])
+                i += 1
+        elif args[i] == "-f":
+            i += 1
+            if i < len(args):
+                file_name = args[i]
+                i += 1
+        else:
+            # Ignore unknown arguments
+            i += 1
```
MNDonut, line 66: Let's find indexes using `.index()` method

app/create_file.py
```diff
@@ -1 +1,91 @@
...
+        path = os.path.join(dir_path, file_name)
+        lines = collect_content_lines()
+        write_content_to_file(path, lines)
+        print(f"Directory and file created/updated: {path}")
+        return
+
+    print("Usage:")
+    print(" python create_file.py -d dir1 dir2")
+    print(" python create_file.py -f filename.txt")
+    print(" python create_file.py -d dir1 dir2 -f filename.txt")
```
MNDonut, line 87: and let's remove these prints as it isn't a part of the task

app/create_file.py
MNDonut: And let's split this logic by functions.
For example a function that accepts a flag that it has to find and returns index or None depending on flag was found or not

app/create_file.py
```diff
@@ -1 +1,91 @@
...
+        print(f"File created/updated: {path}")
+
+    if directory_path and file_name:
+        dir_path = create_path(directory_path)
+        make_directories((dir_path))
+        path = os.path.join(dir_path, file_name)
+        lines = collect_content_lines()
+        write_content_to_file(path, lines)
+        print(f"Directory and file created/updated: {path}")
+        return
```
MNDonut, line 82: The same here. Let's create a function that creates directories (optionally) and a file with the necessary content

MNDonut (changes requested):

MNDonut (approved):

PR #1380, opened 2025-07-14
my-solution

Orixara:


Plovuezz (approved):
I am not sure about "a" in open(), because we should create new file. Also i think you dont need to create list of lines. You just can format each input like f"{count} {line}\n" and just write it to file. But overall it looks ok. Good Job!

PR #1385, opened 2025-07-15
Solution

Hetmarum:


yagginov (approved):
everything is fine, but if you have time, look into what the argparse module is, because it really helps a lot when working with arguments

PR #1386, opened 2025-07-15
Solution

yagginov:


Artcrafterrra (approved):
Good job, using argparse is a greate solution!

PR #1388, opened 2025-07-15
solution py create file from terminal

PavelSemenikhin:


Vskesha (approved):
This script creates a file and writes user input as required. Code is clear and meets all the requirements. Good work.

PR #1393, opened 2025-07-16
create file from terminal

SerhiiBasok:


Sashoxxx (approved):
Good job . You can change "IFs" to "ELIF-ELSE" to make it more proper. Also i guess, there is a mistyping in func name "create_patch"

PR #1395, opened 2025-07-16
solution

Sashoxxx:


app/create_file.py
```diff
@@ -1 +1,49 @@
...
+
+    elif sys.argv[1] == "-f":
+        filename = sys.argv[2]
+
+    else:
+        return
+
+    with open(filename, "a+") as f:
+        while True:
+            counter = 0
```
VadymKyrylko, line 28: You need to take the counter outside the loop.

VadymKyrylko (changes requested):
You need to take the counter outside the loop.

PR #1398, opened 2025-07-16
Solution

Kostenikov:


app/create_file.py
```diff
@@ -1 +1,56 @@
...
+        if new_line:
+            result_data.append(new_line)
+        new_line = input("Enter content line: ")
+
+    file_exist = os.path.exists(filedir)
+
+    with open(filedir, "a") as file:
+        page_number = 1
+
+        if file_exist:
```
MorsImmortalis05, line 29: is this really needed? if you already have with open it means file exists

Kostenikov:
@mateacademy-ai-mentor checklist and reamde doesn't have any information about error massage when filename non provided.

PR #1402, opened 2025-07-16
Solution

reinvvay:


app/create_file.py
```diff
@@ -1 +1,37 @@
-# write your code here
+import os
+import sys
+from datetime import datetime
+
+
+data = sys.argv
+
+path = ""
```
trellle, line 8: u can initialize path with os.getcwd()

PR #1441, opened 2025-08-21
Solution

Konstantin-Svt:


Konstantin-Svt:
@mateacademy-ai-mentor I can't understrand your logic regarding this statement: "-d and -f can appear in any order, possibly multiple times, extract their arguments reliably." Could you clarify, in which directory file2.txt should be created if we have flag line like this: "-f file1.txt -d dir1 -f file2.txt -d dir2". Should it be created in dir1 together with file1.txt and dir2 just empty directory or file2.txt should be in dir2.  What is priority on this. I didn`t refactor code and re-requested review just to get answer from you.

PR #1498, opened 2025-11-01
Solution

Vorongor:


app/create_file.py
```diff
@@ -1 +1,89 @@
...
+            i += 1
+
+    # Get file name
+    if "-f" in command_line:
+        i = command_line.index("-f") + 1
+        if i < len(command_line):
+            file_name = command_line[i]
+
+    # Create directories
+    if dirs:
```
DonDmytron: Your function has a lot of logic, try breaking it down into separate functions: create a file, create directories, and a function that will run it all.

PR #1517, opened 2025-11-12
task

ostapshpeha:


app/create_file.py
```diff
@@ -1 +1,55 @@
...
+    lines.append(f"{count} {line}")
+    count += 1
+
+current_date = datetime.datetime.now()
+header = current_date.strftime("%Y-%m-%d %H:%M:%S")
+body = "\n".join(lines)
+
+if os.path.exists(full_path) and os.path.getsize(full_path) > 0:
+    prefix = "\n\n"
+else:
```
leemonch1k-space: According to checklist[#1] we shouldn't use "else" condition if program will work correct without it

app/create_file.py
```diff
@@ -1 +1,55 @@
...
+            list_dirs.append(args[i])
+            i += 1
+
+    elif args[i] == "-f":
+        if i + 1 < len(args):
+            filename = args[i + 1]
+            i += 2
+        else:
+            raise (Exception("No filename given"))
+    else:
```
leemonch1k-space: According to checklist[#1] we shouldn't use "else" condition if program will work correct without it

leemonch1k-space (changes requested):
Looks good, just remove the "else" condition, and then it should be approved

app/create_file.py
petrykivd: Move path = "." before your if statement

app/create_file.py
```diff
@@ -1 +1,84 @@
-# write your code here
+import os
+import sys
+import datetime
+from typing import Any, LiteralString
```
petrykivd: Why you use LiteralString here?

app/create_file.py
```diff
@@ -1 +1,84 @@
...
+
+        elif args[i] == "-f":
+            if i + 1 < len(args):
+                filename = args[i + 1]
+                i += 2
+            else:
+                raise (Exception("No filename given"))
+        else:
+            i = i + 1
+    return list_dirs, filename
```
petrykivd: This else branch does nothing useful and makes the code harder to follow. If you're already handling -d and -f cases, what exactly is this else supposed to catch? Empty iteration.

app/create_file.py
```diff
@@ -1 +1,84 @@
...
+            while i < len(args) and args[i] not in ("-d", "-f"):
+                list_dirs.append(args[i])
+                i += 1
+
+        elif args[i] == "-f":
+            if i + 1 < len(args):
+                filename = args[i + 1]
+                i += 2
+            else:
+                raise (Exception("No filename given"))
```
petrykivd: Don't wrap Exception in parentheses - it's not a function call. Also, this should probably be ValueError or SystemExit, not generic Exception.

petrykivd (changes requested):

PR #1518, opened 2025-11-12
Solution

leemonch1k-space:


app/create_file.py
```diff
@@ -1 +1,47 @@
...
+
+flag_d_index = args.index("-d") if "-d" in args else None
+flag_f_index = args.index("-f") if "-f" in args else None
+
+
+if flag_d_index is not None:
+    path_parts = []
+    for arg in args[flag_d_index + 1 :]:
+        if arg.startswith("-"):
+            break
```
vakt159: can break if directory starts with - like: -asd
can just make slice of args[flag_d_index + 1: flag_f_index]

leemonch1k-space: You know, i thought about your advice, and yeah, it makes sense

PR #1626, opened 2026-02-06
solution

mykolamosiienko:


app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    f_index = command.index("-f")
+    file_name = command[f_index + 1]
+    d_index = command.index("-d")
+    if d_index > f_index:
+        p_dir = command[d_index + 1 :]
+    else:
+        p_dir = command[d_index + 1 : f_index]
+    dir_path = os.path.join(*p_dir)
+    dir_path_with_file = os.path.join(dir_path, file_name)
+    os.makedirs(dir_path, exist_ok=True)
```
vsmutok: Put this code into a function.

app/create_file.py
```diff
@@ -1 +1,45 @@
...
+    dir_path = os.path.join(*dir_parts)
+    os.makedirs(dir_path, exist_ok=True)
+
+if "-f" in command and "-d" not in command:
+    f_index = command.index("-f")
+    file_name = command[f_index + 1]
+
+    make_file(file_name)
+
+if "-f" in command and "-d" in command:
```
vsmutok: Here you’re doing extra checks. You only need to check once for -f and once for -d; there’s no need to check multiple times.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1664, opened 2026-03-02
Solve the task!

Redum9:


Redum9 (commented):
i solved this proble, check again.

PR #1714, opened 2026-04-07
create function which create file fromterminal

pawunder1992:


app/create_file.py
```diff
@@ -1 +1,41 @@
...
+
+
+def check_for_f(input_data: list) -> None:
+    if not input_data:
+        print("Введіть коректну назву файлу")
+        return
+    if "-f" not in input_data:
+        print("Введіть правильну команду")
+        return
+    path = ("").join(input_data[-1])
```
VadimPusiak788: Unnecessary "".join() on a string. input_data[-1] is already a string

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+import sys
+
+
+def check_for_d(input_data: list) -> None:
+    if "-d" in input_data:
+        new_folder = input_data[input_data.index("-d")
+                                + 1:input_data.index("-f")]
+        path = os.path.join(*new_folder)
+        os.makedirs(path, exist_ok=True)
+        create_file(os.path.join(*new_folder + [input_data[-1]]))
```
VadimPusiak788, line 13: Extract the filename explicitly from the argument that follows "-f" instead of using input_data[-1]. Relying on the last list element is fragile and may produce incorrect results if the argument order changes.

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+        return
+    if "-f" not in input_data:
+        print("Введіть правильну команду")
+        return
+    path = ("").join(input_data[-1])
+    create_file(path)
+
+
+def create_file(path: str) -> None:
+    date_log = str(datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"))
```
VadimPusiak788: strftime already returns a str

app/create_file.py
```diff
@@ -1 +1,41 @@
...
+    with open(path, "a") as file:
+        file.write(date_log)
+        while True:
+            new_line = input("Enter content line: ")
+            if new_line == "stop":
+                break
+            file.write(f"\n{new_line}")
+
+
+check_for_d(sys.argv)
```
VadimPusiak788: if __name__ == "__main__".  Wrap the entry call to prevent execution on import

VadimPusiak788 (changes requested):

app/create_file.py
pawunder1992: solved

app/create_file.py
pawunder1992: done

app/create_file.py
pawunder1992: solved

app/create_file.py
pawunder1992: i added more checkings now

VadimPusiak788 (approved):

PR #1731, opened 2026-04-17
My solution-for-py-create-file-from-terminal

zubak-vasil:


vova-dev-python (approved):
This code is functional and covers the main requirements, but there are a few points where it could be more robust or cleaner according to the guidelines. The logic for end_d on line 20 is clever, but the script might crash if -f is provided without a filename after it (line 24: args[f_idx + 1]). Adding a check to ensure f_idx + 1 < len(args) would make it safer. The variable names like d_idx, f_idx, and f are a bit short. According to the "Clean Code" section of the checklist, descriptive names like directory_index or destination_file are preferred.

AnyoneClown (changes requested):
Refactor with your own solution

PR #1773, opened 2026-05-08
Solução

Raffaele-Maccarrone:


Raffaele-Maccarrone:
@mateacademy-ai-mentor
The fix you proposed—using f"{current_time}\n"—is already exactly what is implemented in the create_file.py code. The hardcoded string in Chinese you mentioned is not present in my file; the implementation already uses the dynamic current_time variable to ensure the timestamp is correctly generated at each execution.

PR #1790, opened 2026-05-27
Solution

swirtam:


swirtam:
@mateacademy-ai-mentor I modified annotation with Union and Optional. I put the code for making catalogues into a separate function for better organization. I also optimized the function from writing content into a file. In the previous solution starting a counter from 0 was intentional and met the requirements. The first line to be written into the file in the loop was a timestamp - this was a 0 line and the condition was added to check if the counter should be added to the line before writing. So the lines inputed by the user were numbered from 1 as required.

PR #1792, opened 2026-05-28
'Solution'

dem1hh:


app/create_file.py
```diff
@@ -1 +1,73 @@
...
+    for arg in args[1:]:
+        if arg == "-d":
+            mode = "dir"
+        elif arg == "-f":
+            mode = "file"
+        else:
+            if mode == "dir":
+                dirs.append(arg)
+            elif mode == "file":
+                file_name = arg
```
vsmutok: Do not use a loop here.
You can check `if "-d" in args` instead....

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+    need_blank_line = False
+    if os.path.exists(final_path):
+        if os.path.getsize(final_path) > 0:
+            need_blank_line = True
+
+    with open(final_path, "a") as f:
+        if need_blank_line:
+            f.write("\n")
+
+        f.write(current_time + "\n")
```
vsmutok:
```suggestion
    with open(final_path, "a") as f:
        if os.path.exists(final_path) and os.path.getsize(final_path) > 0:
            f.write("\n")
    
        f.write(f"{current_time}\n")
```

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+    return dirs, file_name
+
+
+def get_user_content() -> Any:
+    lines = []
+    while True:
+        line = input("Enter content line: ")
+        if line == "stop":
+            break
+        lines.append(line)
```
vsmutok: Consider adding line numbers here right away.

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+            need_blank_line = True
+
+    with open(final_path, "a") as f:
+        if need_blank_line:
+            f.write("\n")
+
+        f.write(current_time + "\n")
+
+        for index, content in enumerate(lines, start=1):
+            f.write(f"{index} {content}\n")
```

app/create_file.py
```diff
@@ -1 +1,73 @@
...
+        if dir_path:
+            final_path = os.path.join(dir_path, file_name)
+        else:
+            final_path = file_name
+
+        user_lines = get_user_content()
+        write_to_file(final_path, user_lines)
+
+
+main()
```
vsmutok, line 67: use 
```
if __name__ == "__main__":
```

vsmutok (changes requested):

app/create_file.py
```diff
@@ -1 +1,67 @@
...
+        current_time = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+        with open(final_path, "a") as f:
+            if os.path.exists(final_path) and os.path.getsize(final_path) > 0:
+                f.write("\n")
+
+            f.write(f"{current_time}\n")
+            f.writelines(user_lines)
+
+
+main()
```
vsmutok, line 67: use 
```
if __name__ == "__main__":
```

vsmutok (changes requested):

PR #1862, opened 2026-07-26
solution

neokitoss:


app/create_file.py
```diff
@@ -1 +1,76 @@
...
+        elif args[i] == "-d":
+            i += 1
+            while i < len(args) and not args[i].startswith("-"):
+                directories.append(args[i])
+                i += 1
+
+        else:
+            i += 1
+
+    return directories, filename
```
vsmutok: Try to avoid using a loop here. You can use `argparse`, or simply check whether the argument is `-f` or `-d`.

app/create_file.py
```diff
@@ -1 +1,76 @@
...
+
+
+def create_directories(directories: list) -> list:
```
vsmutok: Incorrect return type annotation.

app/create_file.py
```diff
@@ -1 +1,76 @@
...
+
+def write_to_file(file_path: str, content_lines: list) -> None:
+    timestamp = datetime.now().strftime("%Y-%m-%d %H:%M:%S")
+    file_exist = os.path.exists(file_path) and os.path.getsize(file_path) > 0
+    with open(file_path, "a") as f:
+        if file_exist:
+            f.write("\n")
+        f.write(timestamp + "\n")
+        for index, line in enumerate(content_lines, 1):
+            f.write(f"{index} {line}\n")
```
You can store the line numbers when the user enters the data. This way, you won't need to iterate over the data twice.

vsmutok (changes requested):

hladiivolodymyr (approved):

PR #1867, opened 2026-07-29
fix:code

EldarRasulov:


app/create_file.py
```diff
@@ -1 +1,57 @@
...
+    mode = "a" if os.path.exists(file_path) else "w"
+
+    with open(file_path, mode, encoding="utf-8") as file:
+        if mode == "a":
+            file.write("\n")
+
+        file.write(f"{timestamp}\n")
+
+        for number, line in enumerate(lines, start=1):
+            file.write(f"{number} {line}\n")
```
Add line numbers to the content while the user is entering phrases.

vsmutok (changes requested):

Sn1kls (approved):
Good work, approve!

